PR #4, opened 2018-09-12
CSS to Less Cart Page

elasticSzyszko:
for review, not ready for merge

### DIFF
--- a/src/containers/CartPage.less
+++ b/src/containers/CartPage.less
@@ -1,0 +1,104 @@
+/**
+ * Copyright © 2018 Elastic Path Software Inc. All rights reserved.
+ *
+ * This is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this license. If not, see
+ *
+ *     https://www.gnu.org/licenses/
+ *
+ *
+ */
+
+ @import './../style/common.less';
+
+.cart-container {
+  min-height: 100%;
+  margin-bottom: 0px;
+
+  .cart-main-container {
+    float: left;
+    width: 100%;
+    position: relative;
+    min-height: 1px;
+    padding-left: 15px;
+    padding-right: 15px;
+    margin-bottom: 2em;
+
+    @media (min-width: @tabletWidth) {
+      float: left;
+      width: 66.66666666666666%;
+    }
+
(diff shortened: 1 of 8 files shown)


src/containers/CartPage.less
```diff
@@ -0,0 +1,275 @@
...
+ *
+ *
+ */
+
+ @import './../style/common.less';
+
+.cart-container{
+  min-height: 100%;
+  margin-bottom: 0px;
+  :after {
```
dusanradovanovic: Let's separate rules with one empty line.

dusanradovanovic: You are missing an `&`. You need to write `&:after` in order to define `:after` pseudoclass for `.cart-container`. That being said, if it looks as expected without it you might as well remove it as it is not being applied right now.

src/containers/CartPage.less
```diff
@@ -0,0 +1,275 @@
...
+    content: "";
+    display: block;
+  }
+  
+  @media screen and (max-width: 768px) {
```
dusanradovanovic: You want import `common.less` and use constants from that file as `@media (min-width: @mobileWidth) {`. Take a look at `addressform.main.less`.

Also you want to add `- 1px` for all `max-width` media queries. So final line would be:
`@media screen and (max-width: @mobileWidth - 1px) {`

src/containers/CartPage.less
```diff
@@ -0,0 +1,275 @@
...
+      display: block;
+      width: 100%;
+    }
+    table tbody tr {
+      display: block;
+      width: 100%;
+    }
+    table tbody tr td {
+      display: block!important;
+      width: 100%!important;
```
dusanradovanovic: Same as previous, avoid `!important`, if you use it put a space before it, if it looks ok the way it's written right now, try to remove the whole line. If you can remove this and previous line, then remove the whole `td` part.

src/containers/CartPage.less
```diff
@@ -0,0 +1,275 @@
...
+ * along with this license. If not, see
+ *
+ *     https://www.gnu.org/licenses/
+ *
+ *
+ */
+
+ @import './../style/common.less';
+
+.cart-container{
```
dusanradovanovic: Space before `{`

src/containers/CartPage.less
```diff
@@ -0,0 +1,275 @@
...
+    }
+  }
+
+  table > tbody > tr:nth-child(odd) > td,
+  table > tbody > tr:nth-child(odd) > th {
+    background-color: #e9e9e9;
+  }
+
+  .cart-container-inner{
+    :before{
```
dusanradovanovic: `&:before`

src/style/reset.less
```diff
@@ -47,9 +47,14 @@ footer, header, hgroup, menu, nav, section {
   display: block;
 }
 
-body {
+body { //this will get overridden by bootstrap?
```
dusanradovanovic: The order styles get included is (take a look at index.js):
- reset.less
- bootstrap
- style.less

So anything you want to keep defined on a global level goes to `style.less`.  `reset.less` merely resets the default browser styles.

elasticSzyszko:
@dusanradovanovic ready for another review, thanks!

PR #21, opened 2018-09-18
Converted application header/footer to use flex

dusanradovanovic:
With this change footer will stay sticky at the bottom of the page even when there is not enough content to fill the whole page.


shaunmaharaj:
I'm currently not seeing the footer stay at the bottom of the page in the lack of content. Most evident on the Cart Page
image

shaunmaharaj:
Spoke to @dusanradovanovic. A future commit will enable the updated styles.

PR #39, opened 2018-10-01
Fixed header structure

shaunmaharaj:


shaunmaharaj:
FYI @sminnath @asmurthwaite

PR #43, opened 2018-10-02
Pdp ui updates

shaunmaharaj:


shaunmaharaj:
@dusanradovanovic @elasticSzyszko @epmkelly I think some of the styles such as fonts need to get brought out into common, so we'll need another round to sync up.

PR #51, opened 2018-10-04
Header redesign and address/payment forms

dusanradovanovic:
Description:

- Redesign of mobile header
- Redesign of new address form 
- Redesign of new payment form 

Tests:

- [x] Manual tests

Documentation:


dusanradovanovic:
image

image

dusanradovanovic:
> Would you happen to know why the nav bar doesn't collapse after performing a search?
Yeah that is pretty annoying. It doesn't collapse because nothing is causing it to collapse :) I'll figure some workaround for that.

PR #60, opened 2018-10-04
Minor footer fixes

dusanradovanovic:
Description:
Minor fixes

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/src/components/appfooter.main.jsx
+++ b/src/components/appfooter.main.jsx
@@ -55,7 +55,7 @@ function AppFooterMain() {
         </div>
         <div className="footer-column social">
           <div className="title">
-            {intl.get('vestri')}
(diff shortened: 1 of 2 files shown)


src/components/appfooter.main.jsx
```diff
@@ -55,7 +55,7 @@ function AppFooterMain() {
         </div>
         <div className="footer-column social">
           <div className="title">
-            {intl.get('vestri')}
+            {intl.get('find-us-online')}
```
shaunmaharaj: Are we going to have the same title for both columns?

dusanradovanovic: No, it's three titles for three columns. There are two different "Find Us Online" columns but only one is visible at any time.

PR #96, opened 2018-10-30
Add dynamic cart item modifier inputs

CPinelli:
Description:
Added dynamic Cart Item Modifier inputs on the cart page and the product details page

Linting:
- [x] No linting errors

Tests:
- [x] Manual tests

Documentation:
It is required to have a configuration object in ep.config.json that contains a "prefix" attribute that is the one used in the cart item modifier object.
For example: If you add a new item modifier input that is called "test", you will most likely give it the name "VESTRI_test" if you are using the Vestri store. Then you should have the following: 
```
"configuration": {
    "prefix": "VESTRI_"
}
```

### DIFF
--- a/src/components/cart.lineitem.jsx
+++ b/src/components/cart.lineitem.jsx
@@ -30,9 +30,6 @@ import {
 import imgPlaceholder from '../images/img-placeholder.png';
 import cortexFetch from '../utils/Cortex';
 import './cart.lineitem.less';
-import epConfig from '../ep.config.json';
-
-const configurationPrefix = epConfig.cartItemModifier.prefix;
 
 const Config = require('Config');
 
@@ -149,9 +146,9 @@ class CartLineItem extends React.Component {
     const keys = Object.keys(item.configuration);
     if (keys) {
       return keys.map(key => (
-        <li className="configuration" key={key.split(configurationPrefix)[1]}>
+        <li className="configuration" key={key}>
           <label htmlFor="option-name" className="option-name">
-            {key.split(configurationPrefix)[1]}
+            {key}
             :&nbsp;
           </label>
           <span>
--- a/src/components/cart.lineitem.jsx
+++ b/src/components/cart.lineitem.jsx
@@ -30,6 +30,9 @@ import {
 import imgPlaceholder from '../images/img-placeholder.png';
 import cortexFetch from '../utils/Cortex';
 import './cart.lineitem.less';
+import epConfig from '../ep.config.json';
+
+const configurationPrefix = epConfig.cartItemModifier.prefix;
 
 const Config = require('Config');
 
@@ -141,6 +144,25 @@ class CartLineItem extends React.Component {
     );
   }
 
+  renderConfiguration() {
+    const { item } = this.props;
+    const keys = Object.keys(item.configuration);
+    if (keys) {
+      return keys.map(key => (
+        <li className="configuration" key={key.split(configurationPrefix)[1]}>
+          <label className="option-name">
+            {key.split(configurationPrefix)[1]}
+            :&nbsp;
+          </label>
+          <span>
+            {item.configuration[key]}
+          </span>
+        </li>
+      ));
+    }
+    return null;
+  }
+
   renderOptions() {
     const { item } = this.props;
     const options = item._item[0]._definition[0]._options;
@@ -197,6 +219,7 @@ class CartLineItem extends React.Component {
         <div className="options-col">
           <ul className="options-container">
             {this.renderOptions()}
+            {this.renderConfiguration()}
           </ul>
         </div>
         <div className="availability-col" data-region="cartLineitemAvailabilityRegion">
--- a/src/components/cart.lineitem.jsx
+++ b/src/components/cart.lineitem.jsx
@@ -150,7 +150,7 @@ class CartLineItem extends React.Component {
     if (keys) {
       return keys.map(key => (
         <li className="configuration" key={key.split(configurationPrefix)[1]}>
-          <label className="option-name">
+          <label htmlFor="option-name" className="option-name">
             {key.split(configurationPrefix)[1]}
             :&nbsp;
           </label>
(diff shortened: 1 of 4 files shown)


src/components/productdisplayitem.main.less
```diff
@@ -102,6 +102,10 @@
             background-position: 152px;
             background-size: 16px 9px;
           }
+
+          .form-control-text {
+            background: #fff;
+          }
```
CPinelli: This is required to override the down arrow that is used for other inputs

src/ep.config.json
```diff
@@ -33,5 +33,8 @@
   "arKit": {
     "enable": true,
     "skuArImagesUrl": "https://s3.amazonaws.com/referenceexp/ar/%sku%.usdz"
+  },
+  "cartItemModifier": {
+    "prefix": "SIG_"
```
CPinelli: Default example with the SIGNATUR store

src/components/productdisplayitem.main.jsx
```diff
@@ -446,6 +477,9 @@ class ProductDisplayItemMain extends React.Component {
                         </option>
                       </select>
                     </div>
+                    {
+                      (isLoading) ? (<div className="miniLoader" />) : ''
+                    }
```
CPinelli: I moved this there because I believe it looks better after I added the cart item modifiers

src/components/cart.lineitem.jsx
```diff
@@ -141,6 +144,25 @@ class CartLineItem extends React.Component {
     );
   }
 
+  renderConfiguration() {
+    const { item } = this.props;
+    const keys = Object.keys(item.configuration);
+    if (keys) {
+      return keys.map(key => (
+        <li className="configuration" key={key.split(configurationPrefix)[1]}>
+          <label className="option-name">
```
shaunmaharaj: Linting error:
`153:11  error  Form label must have ALL of the following types of associated control: id  jsx-a11y/label-has-for`

Need to add the following attribute to the element:
`htmlFor="option-name"`

shaunmaharaj:
Confirmed passes AUT successfully. Proceeding with review.

src/components/cart.lineitem.jsx
```diff
@@ -30,6 +30,9 @@ import {
 import imgPlaceholder from '../images/img-placeholder.png';
 import cortexFetch from '../utils/Cortex';
 import './cart.lineitem.less';
+import epConfig from '../ep.config.json';
```
shaunmaharaj: This is not necessary as it's already loaded below:
`const Config = require('Config');`

src/components/cart.lineitem.jsx
```diff
@@ -30,6 +30,9 @@ import {
...
+
+const configurationPrefix = epConfig.cartItemModifier.prefix;
```
shaunmaharaj: Not necessary as you can reuse the existing config and pull what you need later in your block with:
`Config.cartItemModifier.prefix`

src/components/productdisplayitem.main.jsx
```diff
@@ -31,6 +31,7 @@ import {
 import imgPlaceholder from '../images/img-placeholder.png';
 import ProductRecommendationsDisplayMain from './productrecommendations.main';
 import cortexFetch from '../utils/Cortex';
+import epConfig from '../ep.config.json';
```
shaunmaharaj: This is not necessary as it's already loaded below:
`const Config = require('Config');`

src/components/productdisplayitem.main.jsx
```diff
@@ -63,6 +64,8 @@ const zoomArray = [
   'code',
 ];
 
+const configurationPrefix = epConfig.cartItemModifier.prefix;
```
shaunmaharaj: Not necessary as you can reuse the existing config and pull what you need later in your block with:
`Config.cartItemModifier.prefix`

src/components/productdisplayitem.main.jsx
```diff
@@ -163,6 +168,15 @@ class ProductDisplayItemMain extends React.Component {
     this.setState({ itemQuantity: parseInt(event.target.value, 10) });
   }
 
+  handleConfiguration(configuration, event) {
+    let { itemConfiguration } = this.state;
+    if (!itemConfiguration) {
```
shaunmaharaj: Is there ever a point where `itemConfiguration ` would need this check? Looks like your initial value for it in the state is an empty set, which you are fetching.

src/components/productdisplayitem.main.jsx
```diff
@@ -207,6 +221,7 @@ class ProductDisplayItemMain extends React.Component {
           },
           body: JSON.stringify({
             quantity: itemQuantity,
+            configuration: itemConfiguration,
```
shaunmaharaj: This will get passed into the request body even if it's empty.

src/components/cart.lineitem.jsx
```diff
@@ -141,6 +144,25 @@ class CartLineItem extends React.Component {
     );
   }
 
+  renderConfiguration() {
+    const { item } = this.props;
+    const keys = Object.keys(item.configuration);
+    if (keys) {
+      return keys.map(key => (
+        <li className="configuration" key={key.split(configurationPrefix)[1]}>
```
shaunmaharaj: Trying this out. Why is it necessary to even provide a configurationPrefix? It's sufficient enough to allow this to be set through CM and display the same value in the storefront as below.
```suggestion
        <li className="configuration" key={key}>
        <label htmlFor="option-name" className="option-name">
            {key}
        </label>
```
image

CPinelli: You are right, when Mark created the modifier input, he added `SIG_` as a prefix, so I figured out it was like a convention within EP. I should've asked, I fixed it!

src/components/productdisplayitem.main.jsx
CPinelli: Fixed

src/components/productdisplayitem.main.jsx
CPinelli: fixed, indeed it cannot happen, I did that on the demo web app, but here I did it properly lol

src/components/productdisplayitem.main.jsx
CPinelli: Removed

src/components/productdisplayitem.main.jsx
CPinelli: Removed

src/components/cart.lineitem.jsx
CPinelli: Removed

PR #135, opened 2018-12-06
Display lineitem promotions

maennis-ep:
Description:
Cart lineitems now show the values from the `appliedpromotions` resource.  Also, changed the cart applied promotions to display a promotions `name` if it does not have a `display-name`, otherwise a blank line is all that appears.

Linting:
- [x] No linting errors

Tests:
- [x] Manual tests

To test the changes manually, add a Cart Promotion that discounts a certain product when it is in your cart.  Add that item to the cart, and you will see the promotion name in the upper right hand corner of the cart line item

Documentation:

### DIFF
--- a/src/components/cart.lineitem.less
+++ b/src/components/cart.lineitem.less
@@ -61,13 +61,9 @@
   }
 
   .promotions-col {
-    display: inline-block;
+    display: block;
     width: 33%;
-    text-align: right;
-    @media (max-width: @mobileWidth - 1px) {
-      width: 100%;
-      text-align: left;
-    }
+    text-align: left;
   }
   
   .availability-col {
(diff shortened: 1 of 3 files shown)


src/components/cart.lineitem.less
```diff
@@ -51,13 +52,24 @@
...
+    vertical-align: top;
   }
 
   .options-col {
     display: inline-block;
     width: 50%;
   }
 
+  .promotions-col {
+    display: inline-block;
```
shaunmaharaj: I'd prefer keeping this fixed to the left of the container as below. 
image

It's more mobile-friendly and wraps better dynamically in our containers. The current styles will break on mobile-tablet views. Also as the length of the product name increases, it will overlap with the width of this promotions div.

src/components/cart.lineitem.less
```diff
@@ -51,13 +52,24 @@
...
+    width: 33%;
+    text-align: right;
```
shaunmaharaj: To achieve the styles in the screenshot above:
```suggestion
    display: block;
    width: 33%;
    text-align: left;
```

PR #136, opened 2018-12-07
Intoli script modifications to fit our Jenkins Node OS

epmkelly:
Description:
We're using a install-google-chrome script provided by intoli (https://intoli.com/blog/installing-google-chrome-on-centos/)
This script has allowed us to install google chrome onto our jenkins nodes, which are using Amazon Linux AMI 2017.03
The script helps to install needed dependencies from other os repositories, and loops to find any other needed dependencies for those dependencies. This is needed because chrome isn't actually supported on the OS that we're using. This script has helped us to work around those dependencies, but it was currently breaking as one of the dependencies couldn't be found. We've added checks for those dependencies and included the mirror of the last working version we had.

The pipeline is passing now with these changes, but we should look into and consider a Jenkins node for our pipeline that has a simpler install for google-chrome.

image


src/ep.config.json
```diff
@@ -2,7 +2,7 @@
   "cortexApi": {
     "path": "/cortex",
     "scope": "vestri",
-    "pathForProxy": "http://localhost:9080"
+    "pathForProxy": "http://vestri.us.epdemos.com"
```
shaunmaharaj: Can we revert this? OOTB we should be on `localhost:9080`

src/components/appheader.main.jsx
```diff
@@ -92,7 +92,9 @@ class AppHeaderMain extends React.Component {
           </div>
 
           <div className="cart-link-container">
-            <Link className="cart-link" to="/mybag" />
+            <Link className="cart-link" to="/mybag">
```
shaunmaharaj: I threw this in since the new version of Chrome we're using wasn't displaying our CSS when the UT was run. If that's still the case, this can stay as it's a good fallback to have text in place.

src/ep.config.json
epmkelly: yeah, good catch. I didn't mean to put this in. was leftover in my workspace

src/components/appheader.main.less
```diff
@@ -90,6 +90,8 @@
         height: 24px;
         background: #ffffff00 url(../images/header-icons/bag-icon.png) no-repeat 0px;
         background-position: center top;
+        line-height: 0;
```
shaunmaharaj: If you're removing the cart-link change, this should be removed too

PR #170, opened 2019-01-04
Added shareThis react component and buttons on product page

skarpyak:
Description:

Linting:

- [x] No linting errors

Tests:

- [ ] Manual tests

Documentation:

### DIFF
--- a/src/components/productdisplayitem.main.jsx
+++ b/src/components/productdisplayitem.main.jsx
@@ -422,7 +422,7 @@ class ProductDisplayItemMain extends React.Component {
       let productLink = '';
       if (productData._availability.length >= 0) {
         if (productData._code) {
-          productLink = `https://reference.elasticpath.com/itemdetail/${productData._code[0].code}`;
+          productLink = `${window.location.origin}/itemdetail/${productData._code[0].code}`;
         }
         if (productData._availability[0].state === 'AVAILABLE') {
           availabilityString = intl.get('in-stock');
@@ -435,7 +435,9 @@ class ProductDisplayItemMain extends React.Component {
           availabilityString = intl.get('out-of-stock');
         }
       }
-      const shareImage = Config.skuImagesUrl.replace('%sku%', productData._code[0].code);
+      const productTitle = productData._definition[0]['display-name'];
+      const productDescription = productData._definition[0].details ? (productData._definition[0].details.find(detail => detail['display-name'] === 'Summary')).value : '';
+      const productImage = Config.skuImagesUrl.replace('%sku%', productData._code[0].code);
       const featuredProductAttribute = (productData._definition[0].details) ? (productData._definition[0].details.find(detail => detail['display-name'] === 'Featured')) : '';
       // Set the language-specific configuration for indi integration
       Config.indi.productReview.title = intl.get('indi-product-review-title');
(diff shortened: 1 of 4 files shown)


src/components/productdisplayitem.main.jsx
```diff
@@ -418,7 +419,11 @@ class ProductDisplayItemMain extends React.Component {
       }
       let availability = (productData._addtocartform[0].links.length > 0);
       let availabilityString = '';
+      let productLink = '';
       if (productData._availability.length >= 0) {
+        if (productData._code) {
+          productLink = `https://reference.elasticpath.com/itemdetail/${productData._code[0].code}`;
```
shaunmaharaj: We can't hardcode the domain for this link. Need to use window.location or hostname to grab it.

src/components/productdisplayitem.main.jsx
```diff
@@ -577,6 +583,36 @@ class ProductDisplayItemMain extends React.Component {
...
+                      'email',
+                    ],
+                    padding: 12, // padding within buttons (INTEGER)
+                    radius: 4, // the corner radius on each button (INTEGER)
+                    size: 40, // the size of each button (INTEGER)
+
+                    // OPTIONAL PARAMETERS
+                    url: productLink, // (defaults to current url)
+                    image: shareImage, // (defaults to og:image or twitter:image)
+                    description: 'custom text', // (defaults to og:description or twitter:description)
```
shaunmaharaj: Is it possible to pass the product description directly here?
Available in `productData._definition[0].details`

src/components/productdisplayitem.main.jsx
```diff
@@ -577,6 +583,36 @@ class ProductDisplayItemMain extends React.Component {
...
+                    title: 'Test title', // (defaults to og:title or twitter:title)
```
shaunmaharaj: Can this be updated to the product name?

src/index.html
```diff
@@ -1,142 +1,146 @@
...
+<html manifest="manifest.appcache">
+
+<head>
+
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+  <meta name="description" content="Elastic Path&trade; Visual Cortex - App Framework">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="theme-color" content="#5bc0de" />
+  <meta property="og:image" content="https://s3-us-west-2.amazonaws.com/ep-demo-images/VESTRI_VIRTUAL/VESTRI_MODEL_X_FOR_KIDS.png">
```
shaunmaharaj: Is this intended to be hardcoded?

shaunmaharaj:
Confirmed no linting errors 👍 
Confirmed passes automated unit tests 👍

src/index.html
skarpyak: we can add default share image, now share with Facebook will look like this:
image

shaunmaharaj: Can we use this public image as a placeholder?
https://www.elasticpath.com/sites/default/files/favicon_0.ico

src/index.html
```diff
@@ -1,142 +1,146 @@
...
+  <meta property="og:image:type" content="image/png">
+  <meta property="og:image:width" content="1024">
+  <meta property="og:image:height" content="1024">
```
shaunmaharaj: As of your latest commit, it looks like these `og:image` properties are no longer necessary

skarpyak: we still need it, image option works only for pinterest

skarpyak: also for reach sharing in facebook we need manipulate metatags

src/index.html
skarpyak: there is strong image size requirements, icon image too small:
https://developers.facebook.com/docs/sharing/best-practices/#precaching

skarpyak: To summarize:
Minimum size in pixels is 600x315
Recommended size is 1200x630 - Images this size will get a larger display treatment.

shaunmaharaj: Ah alright, thanks. I'd prefer using a generic image if it has to be hardcoded rather than hardcoding one of the product images directly from S3. 

How about `img-placeholder.png`? or `manifest-images/icons-512.png`?

src/components/productdisplayitem.main.jsx
```diff
@@ -430,6 +435,9 @@ class ProductDisplayItemMain extends React.Component {
           availabilityString = intl.get('out-of-stock');
         }
       }
+      const productTitle = productData._definition[0]['display-name'];
+      const productDescription = productData._definition[0].details ? (productData._definition[0].details.find(detail => detail['display-name'] === 'Summary')).value : '';
```
shaunmaharaj: This doesn't actually fetch the description summary. Take a look at the `renderAttributes()` function in this class for info on grabbing the value.

It's returned as `detail['display-value']` of the `detail` from your find

PR #178, opened 2019-01-10
Nested dynamic bundles

maennis-ep:
Description:
Added nested configurators for the Dynamic Bundles accelerators.  Once a dynamic bundle has been added to cart, the line item will have a configure button that will allow the user to select the bundle constituents.  This PR allows an unlimited number of nested dynamic bundles.

To make use of the Dynamic Bundle functionality the Cortex API the Reference Experience is pointing at must have the Dynamic Bundle accelerator.

Linting:
- [x] No linting errors

Tests:
- [x] Manual tests

1. Create two (or more) Dynamic Bundles through CM
2. Add any child bundles to their parents as constituents
3. Go to the Reference Experience and add the top level bundle to your cart
4. Configure the top level bundle
5. Configure any child bundles
6. Profit

Documentation:


package.json
```diff
@@ -45,6 +45,7 @@
     "react-motion": "^0.4.4",
     "react-onclickoutside": "^5.3.3",
     "react-overlays": "^0.6.6",
+    "react-responsive-modal": "^3.5.1",
```
shaunmaharaj: Good usage of this! I'm going to look into re-using this for the Login modal too 👍

src/components/cart.lineitem.jsx
```diff
@@ -383,6 +499,27 @@ class CartLineItem extends React.Component {
...
+        ) : ('')
+        }
+        {(item._dependentoptions && item._dependentoptions[0] && (item._dependentoptions[0]._element || item._dependentlineitems[0]._element)) ? (
+          <div className="configure-btn-col">
+            <button className="ep-btn primary small btn-cart-configureBundle" type="button" onClick={() => this.handleModalOpen()} data-toggle="modal" data-target={`#bundle-configuration-modal-${itemCodeString}`}>
+              <span className="btn-text">
+                {intl.get('configure-bundle')}
+              </span>
+            </button>
+            <AppModalBundleConfigurationMain key={`app-modal-bundle-configuration-main_${itemCodeString}`} handleModalClose={this.handleModalClose} bundleConfigurationItems={item} openModal={openModal} />
```
shaunmaharaj: Just want to be sure, REACT is successfully keeping state of which modal is loaded into the DOM at a time and their levels?

src/components/productlistitem.main.jsx
```diff
@@ -84,8 +84,8 @@ class ProductListItemMain extends React.Component {
     const { productData } = this.state;
     const { offerData } = this.props;
     if (productData) {
-      let listPrice = 'n/a';
-      let itemPrice = 'n/a';
+      let listPrice = '';
```
shaunmaharaj: What's the text to display if list & item prices are not set?

src/components/appmodalbundleconfiguration.main.jsx
```diff
@@ -0,0 +1,186 @@
...
+                  {intl.get('configure-bundle-configurator')}
+                </h2>
+              </div>
+
+              {(dependantItemData._dependentlineitems[0] && dependantItemData._dependentlineitems[0]._element) ? (
+                <div className="modal-body bundle-configurator-selected-items">
+                  <h2 className="modal-title">
+                    {intl.get('bundle-configurator-selected-items')}
+                  </h2>
+                  <div className="wish-list-main-inner table-responsive">
```
shaunmaharaj: No need for the `wish-list-main-inner` class here

src/components/appmodalbundleconfiguration.main.jsx
```diff
@@ -0,0 +1,186 @@
...
+                </div>
+              ) : ('')
+              }
+
+              {(dependantItemData._dependentoptions[0] && dependantItemData._dependentoptions[0]._element) ? (
+                <div className="modal-body bundle-configurator-available-items">
+                  <h2 className="modal-title">
+                    {intl.get('bundle-configurator-available-items')}
+                  </h2>
+                  <div className="wish-list-main-inner table-responsive">
```
shaunmaharaj: No need for the `wish-list-main-inner` class here

src/components/productlistitem.main.jsx
maennis-ep: I think that this was originally set for the Dynamic Bundle POC because we were not pulling in price for the `dependentlineitems `or `dependentoptions `but that is no longer the case.  I will add the n/a back.

src/components/cart.lineitem.jsx
maennis-ep: It is!  When you start going into the nested modals, all open modals remain in the DOM but only one is displayed.  state changes in any child modal are correctly reflected throughout all the parents.

PR #203, opened 2019-01-19
EAM Login Flow

shaunmaharaj:
Description:

- Incorporate toggles for b2b features
- Initial push for EAM login flow

Linting:

- [x] No linting errors

Tests:

- [ ] Manual tests

Documentation:

### DIFF
--- a/src/components/appmodalcartselect.main.jsx
+++ b/src/components/appmodalcartselect.main.jsx
@@ -25,7 +25,13 @@ import ReactRouterPropTypes from 'react-router-prop-types';
 import intl from 'react-intl-universal';
 import { withRouter } from 'react-router-dom';
 import Modal from 'react-responsive-modal';
+import cortexFetch from '../utils/Cortex';
 import './appmodalcartselect.main.less';
+
+const zoomArray = [
+  'authorizationcontexts',
+  'authorizationcontexts:element',
+];
 
 const Config = require('Config');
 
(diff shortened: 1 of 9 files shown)


src/components/appmodalcartselect.main.jsx
```diff
@@ -48,15 +54,38 @@ class AppModalCartSelectMain extends React.Component {
...
+  fetchOrganizationData() {
+    cortexFetch(`/eam?zoom=${zoomArray.join()}`, {
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: localStorage.getItem(`${Config.cortexApi.scope}_oAuthTokenEam`),
+      },
+    })
+      .then(res => res.json())
+      .then((res) => {
+        const orgEamData = res._authorizationcontexts[0]._element.find(element => element.name === Config.cortexApi.scope.toUpperCase());
```
petro97: If I logged in I still have empty option cart:
screen shot 2019-01-22 at 11 13 49 am

I noticed all will works correctly if use 'MOBEE' instead 'VESTRI'   from our config: res._authorizationcontexts[0]._element.find(element => element.name === 'MOBEE');

I think backend settings should be updated to 'VESTRI'?

shaunmaharaj: @petro97 The API should be returning an organization of the same scope that we have in the storefront. It's similar to how we select the "scope" in the API when doing a storefront shopping call to the backend. In a real b2b flow for the APIs, there would be multiple organizations returned. The storefront would then filter through the list and get the data for the scope which the store is configured for (Our store uses Vestri, but the current data on the backend has Mobee). Let me know if you have more questions on this and we can chat over slack 😃

shaunmaharaj:
Thanks for the review @yulia-dnistrian @petro97. I'll merge this one in and we can continue working off what we have in the codebase.

PR #220, opened 2019-01-25
Btb login flows bc

rostyk-kanafotskyy:
Description:

Linting:

- [x] No linting errors

Tests:

- [ ] Manual tests

Documentation:

### DIFF
--- a/src/components/appheader.main.jsx
+++ b/src/components/appheader.main.jsx
@@ -104,6 +104,7 @@ class AppHeaderMain extends React.Component {
       isOffline, cartData, isLoading, isSearchFocused,
     } = this.state;
     const isInStandaloneMode = window.navigator.standalone;
+    const isB2B = localStorage.getItem(`${Config.cortexApi.scope}_b2bCart`);
     return [
       <header key="app-header" className="app-header">
         <AppHeaderTop />
@@ -127,9 +128,11 @@ class AppHeaderMain extends React.Component {
             </div>
           </div>
 
-          <div className="search-container">
-            <AppHeaderSearchMain isMobileView={false} />
-          </div>
+          {!isB2B && !Config.b2b.enable && (
+            <div className="search-container">
+              <AppHeaderSearchMain isMobileView={false} />
+            </div>
+          )}
 
           <div className="search-toggle-btn-container">
             <button
@@ -149,17 +152,18 @@ class AppHeaderMain extends React.Component {
             <AppHeaderLoginMain isMobileView={false} />
           </div>
 
-          <div className="cart-link-container">
-            <Link className="cart-link" to="/mybag">
-              {cartData && cartData['total-quantity'] !== 0 && !isLoading && (
-                <span className="cart-link-counter">
-                  {cartData['total-quantity']}
-                </span>
-              )}
-              {intl.get('shopping-bag-nav')}
-            </Link>
-          </div>
-
+          {!isB2B && !Config.b2b.enable && (
+            <div className="cart-link-container">
+              <Link className="cart-link" to="/mybag">
+                {cartData && cartData['total-quantity'] !== 0 && !isLoading && (
+                  <span className="cart-link-counter">
+                    {cartData['total-quantity']}
+                  </span>
+                )}
+                {intl.get('shopping-bag-nav')}
+              </Link>
+            </div>
+          )}
 
           <div className="toggle-btn-container">
             {(isInStandaloneMode) ? (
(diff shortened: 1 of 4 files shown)


shaunmaharaj:
Need to either remove “undefined” from profile dropdown in b2b flows, or replace it with a meaningful value.

shaunmaharaj:
Can we please hide functionality that will not be available unless a user logs in as a buyer?
example: search, shopping bag should be hidden from the UI if `b2b.enable` is `true` and the user has not logged in as a buyer.

shaunmaharaj:
No further comments

src/components/appheader.main.jsx
```diff
@@ -104,6 +104,7 @@ class AppHeaderMain extends React.Component {
       isOffline, cartData, isLoading, isSearchFocused,
     } = this.state;
     const isInStandaloneMode = window.navigator.standalone;
+    const isB2B = localStorage.getItem(`${Config.cortexApi.scope}_b2bCart`);
```
shaunmaharaj: I suggest renaming this var to something other than `isB2B` for readability. The current name implies it knows if the flow is B2B, but it really only determines if a cart has been selected

src/components/appheader.main.jsx
```diff
@@ -127,9 +128,11 @@ class AppHeaderMain extends React.Component {
             </div>
           </div>
 
-          <div className="search-container">
-            <AppHeaderSearchMain isMobileView={false} />
-          </div>
+          {!isB2B && !Config.b2b.enable && (
```
shaunmaharaj: This condition won't work as expected. We need to show these components only when we're in B2C, or only in B2B after a buyer has logged in.

Something like the following should work better:
`{!Config.b2b.enable || (Config.b2b.enable && isB2B) && (....`

src/components/appheader.main.jsx
```diff
@@ -149,17 +152,18 @@ class AppHeaderMain extends React.Component {
...
-              {cartData && cartData['total-quantity'] !== 0 && !isLoading && (
-                <span className="cart-link-counter">
-                  {cartData['total-quantity']}
-                </span>
-              )}
-              {intl.get('shopping-bag-nav')}
-            </Link>
-          </div>
-
+          {!isB2B && !Config.b2b.enable && (
```
shaunmaharaj: This condition won't work as expected. We need to show these components only when we're in B2C, or only in B2B after a buyer has logged in.

Something like the following should work better:
`{!Config.b2b.enable || (Config.b2b.enable && isB2B) && (....`

src/components/appheader.main.jsx
```diff
@@ -183,29 +187,34 @@ class AppHeaderMain extends React.Component {
         </div>
 
         <div className="collapsable-container collapse collapsed">
-          <div className="search-container">
-            <AppHeaderSearchMain isMobileView isFocused={isSearchFocused} />
-          </div>
+          {!isB2B && !Config.b2b.enable && (
```
shaunmaharaj: This condition won't work as expected. We need to show these components only when we're in B2C, or only in B2B after a buyer has logged in.

Something like the following should work better:
`{!Config.b2b.enable || (Config.b2b.enable && isB2B) && (....`

src/components/appheader.main.jsx
```diff
@@ -183,29 +187,34 @@ class AppHeaderMain extends React.Component {
...
-              <div data-toggle="collapse" data-target=".collapsable-container">
-                {intl.get('shopping-bag-nav')}
-                <div className="cart-link-counter-container">
-                  {cartData && cartData['total-quantity'] !== 0 && !isLoading && (
-                    <span className="cart-link-counter">
-                      {cartData['total-quantity']}
-                    </span>
-                  )}
+
+          {!isB2B && !Config.b2b.enable && (
```
shaunmaharaj: This condition won't work as expected. We need to show these components only when we're in B2C, or only in B2B after a buyer has logged in.

Something like the following should work better:
`{!Config.b2b.enable || (Config.b2b.enable && isB2B) && (....`

src/components/appmodalcartselect.main.jsx
```diff
@@ -71,48 +62,69 @@ class AppModalCartSelectMain extends React.Component {
...
-      .then((res) => {
-        const orgEamData = res._authorizationcontexts[0]._element.find(element => element.name === Config.cortexApi.scope.toUpperCase());
-        this.setState({
-          orgEamData,
-        });
-      })
-      .catch((error) => {
-        // eslint-disable-next-line no-console
-        console.error(error.message);
+    login()
```
shaunmaharaj: The `login` will return a 404 at this state and never makes it into the `admin_eam` call.
image

PR #240, opened 2019-02-06
Removed the comparison button on the search page

petro97:
Description:

Linting:

- [x] No linting errors

Tests:

- [ ] Manual tests

Documentation:


src/components/productlist.main.jsx
```diff
@@ -132,13 +132,15 @@ class ProductListMain extends React.Component {
...
       return (
         <div className="product-list-container" data-region="categoryBrowseRegion">
-          <div className="compare-button">
-            <button type="button" className="ep-btn primary top-compare-link" onClick={(compareList.length <= 1) ? () => this.handleCompare(isCompare) : ''}>
-              {(compareList.length > 1) ? (
-                <Link className="toggle-compare-link" to={compareLink}>{intl.get('compare-products')}</Link>
-              ) : `${intl.get('compare')}`}
-            </button>
-          </div>
+          {!categoryModel._facets ? (
```
shaunmaharaj: This isn't the best way of determining if we're on the search results page. Facets can be disabled in CM, and won't show up at all if none are available. 
The best alternatives may be to:
• either use the url or type from self in the response object and check if it contains `search`
• either pass a property from `SearchResultsItemsMain` down to the productlist component and use that to check

PR #259, opened 2019-02-26
Power reviews integration

zilkerdev:
Description:

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:


src/components/productdisplayitem.main.jsx
```diff
@@ -460,42 +520,65 @@ class ProductDisplayItemMain extends React.Component {
     );
   }
 
+  renderPowerReviewsElements() {
```
shaunmaharaj: *Suggestion*
This looks like pieces can be brought out into it's own `powerreviews` component and referenced wherever needed. What do you think?

src/containers/WriteReviewPage.jsx
```diff
@@ -0,0 +1,65 @@
...
+ *     https://www.gnu.org/licenses/
+ *
+ *
+ */
+
+import React from 'react';
+import queryString from 'query-string';
+import scriptjs from 'scriptjs';
+import ReactRouterPropTypes from 'react-router-prop-types';
+import '../components/powerreviews.less';
```
shaunmaharaj: If the powerreviews code was in a separate component, the styles would not have to be imported twice and the component can adjust depending on where it's displayed

src/containers/WriteReviewPage.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+  componentDidMount() {
+    const { location } = this.props;
+    const url = location.search;
+    const params = queryString.parse(url);
+
+    const productCode = params.pr_page_id;
+
+    scriptjs(powerReviewsRemoteScriptUrl, () => {
+      // eslint-disable-next-line no-undef
```
shaunmaharaj: *Suggestion*
Had to add this since we're mixing local browser scripts with REACT JS. Might need to find a way around it.

src/components/productdisplayitem.main.jsx
```diff
@@ -345,6 +357,54 @@ class ProductDisplayItemMain extends React.Component {
     this.setState({ selectionValue: event.target.value });
   }
 
+  extractPrice(productData) {
```
shaunmaharaj: Thank you for bringing these out! It's much cleaner 😄

src/components/productdisplayitem.main.jsx
shaunmaharaj: Here's a sample of how it can be done (previously done for another component):
https://github.com/elasticpath/react-pwa-reference-storefront/blob/master/src/components/indirecommendations.main.jsx

src/containers/WriteReviewPage.jsx
```diff
@@ -0,0 +1,65 @@
...
+      POWERREVIEWS.display.render({
+        api_key: Config.PowerReviews.api_key,
+        locale: 'en_US',
```
shaunmaharaj: I notice this is hardcoded. Will it accept the values of our locale's we set in the storefront in localstorage? 
Example: fr-FR, en-CA

PR #260, opened 2019-03-01
Display message for roles with more limited resources

petro97:
Description:

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/src/components/appheader.main.jsx
+++ b/src/components/appheader.main.jsx
@@ -87,10 +87,12 @@ class AppHeaderMain extends React.Component {
       })
         .then(res => res.json())
         .then((res) => {
-          this.setState({
-            cartData: res._defaultcart[0],
-            isLoading: false,
-          });
+          if (res && res._defaultcart) {
+            this.setState({
+              cartData: res._defaultcart[0],
+              isLoading: false,
+            });
+          }
         })
         .catch((error) => {
           // eslint-disable-next-line no-console
@@ -103,6 +105,7 @@ class AppHeaderMain extends React.Component {
     const {
       isOffline, cartData, isLoading, isSearchFocused,
     } = this.state;
+    const availability = !!cartData;
     const isInStandaloneMode = window.navigator.standalone;
     const isB2bCartSelected = localStorage.getItem(`${Config.cortexApi.scope}_b2bCart`);
     return [
@@ -124,35 +127,38 @@ class AppHeaderMain extends React.Component {
 
           <div className="central-container">
             <div className="horizontal-menu">
+              {availability && (
               <AppHeaderNavigationMain isOfflineCheck={this.handleIsOffline} isOffline={isOffline} isMobileView={false} />
+              )}
             </div>
           </div>
 
-          {(!Config.b2b.enable || (Config.b2b.enable && isB2bCartSelected)) && (
+          {(!Config.b2b.enable || (Config.b2b.enable && isB2bCartSelected && availability)) && (
             <div className="search-container">
               <AppHeaderSearchMain isMobileView={false} />
             </div>
           )}
-
-          <div className="search-toggle-btn-container">
-            <button
-              className="search-toggle-btn"
-              type="button"
-              data-toggle="collapse"
-              data-target=".collapsable-container"
-              aria-expanded="false"
-              aria-label="Toggle navigation"
-              onClick={this.handleInputFocus}
-            >
-              <div className="search-icon" />
-            </button>
-          </div>
+          {availability && (
+            <div className="search-toggle-btn-container">
+              <button
+                className="search-toggle-btn"
+                type="button"
+                data-toggle="collapse"
+                data-target=".collapsable-container"
+                aria-expanded="false"
+                aria-label="Toggle navigation"
+                onClick={this.handleInputFocus}
+              >
+                <div className="search-icon" />
+              </button>
+            </div>
+          )}
 
           <div className="login-container">
-            <AppHeaderLoginMain isMobileView={false} />
+            <AppHeaderLoginMain isMobileView={false} permission={availability} />
           </div>
 
-          {(!Config.b2b.enable || (Config.b2b.enable && isB2bCartSelected)) && (
+          {(!Config.b2b.enable || (Config.b2b.enable && isB2bCartSelected && availability)) && (
             <div className="cart-link-container">
               <Link className="cart-link" to="/mybag">
                 {cartData && cartData['total-quantity'] !== 0 && !isLoading && (
@@ -187,7 +193,7 @@ class AppHeaderMain extends React.Component {
         </div>
 
         <div className="collapsable-container collapse collapsed">
-          {(!Config.b2b.enable || (Config.b2b.enable && isB2bCartSelected)) && (
+          {(!Config.b2b.enable || (Config.b2b.enable && isB2bCartSelected && availability)) && (
             <div className="search-container">
               <AppHeaderSearchMain isMobileView isFocused={isSearchFocused} />
             </div>
@@ -196,7 +202,7 @@ class AppHeaderMain extends React.Component {
             <AppHeaderLocaleMain isMobileView />
           </div>
 
-          {(!Config.b2b.enable || (Config.b2b.enable && isB2bCartSelected)) && (
+          {(!Config.b2b.enable || (Config.b2b.enable && isB2bCartSelected && availability)) && (
             <div className="mobile-cart-link-container">
               <Link
                 className="cart-link"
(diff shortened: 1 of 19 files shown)


src/containers/ProfilePage.jsx
```diff
@@ -78,6 +79,11 @@ class ProfilePage extends React.Component {
         })
         .then(res => res.json())
         .then((res) => {
+          if (res.links.length === 0 && !res._defaultprofile) {
+            this.setState({
+              permission: true,
```
shaunmaharaj: Can we rename `permission` to something clearer? Right now it looks like you *do* have permission to view this resource by setting `permission` to true.
Perhaps use `invalidPermission` or keep `permission` and set it to false when you don't have permission?

src/containers/ProfilePage.jsx
```diff
@@ -78,6 +79,11 @@ class ProfilePage extends React.Component {
         })
         .then(res => res.json())
         .then((res) => {
+          if (res.links.length === 0 && !res._defaultprofile) {
```
shaunmaharaj: The check `res.links.length === 0` is not triggered, so the message never gets displayed

src/containers/WishListsPage.jsx
```diff
@@ -81,6 +82,11 @@ class WishListsPage extends React.Component {
       })
         .then(res => res.json())
         .then((res) => {
+          if (res.links.length === 0 && !res._defaultwishlist) {
```
shaunmaharaj: The check `res.links.length === 0` is not triggered, so the message never gets displayed

src/containers/CartPage.jsx
```diff
@@ -92,6 +93,11 @@ class CartPage extends React.Component {
       })
         .then(res => res.json())
         .then((res) => {
+          if (res.links.length === 0 && !res._defaultcart) {
```
shaunmaharaj: The check `res.links.length === 0` is not triggered, so the message never gets displayed

src/containers/CartPage.less
```diff
@@ -50,6 +50,14 @@
   }
 
   .cart-container-inner {
+    .message-permission {
```
shaunmaharaj: This should go in our `common` styles if it's touching multiple components.

src/containers/ProfilePage.less
```diff
@@ -98,4 +98,12 @@
     font-size: 16px;
     font-weight: bold;
   }
+
+  .message-permission {
+    h2{
+      color: red;
```
shaunmaharaj: This should go in our `common` styles if it's touching multiple components.

src/containers/WishListsPage.less
```diff
@@ -50,6 +50,14 @@
   }
 
   .wish-list-container-inner {
+    .message-permission {
```
shaunmaharaj: This should go in our `common` styles if it's touching multiple components.

src/containers/WishListsPage.less
```diff
@@ -50,6 +50,14 @@
...
+      h2{
+        color: red;
+        text-align: center;
```
shaunmaharaj: Can we please also add some top padding so it's not flush with any divs before it
image

src/localization/en-CA.json
```diff
@@ -160,6 +160,7 @@
   "addresses": "Addresses",
   "personal-information": "Personal Information",
   "no-payment-methods-message": "You have no saved paymentMethods.",
+  "permission-message": "You don't have permission to access",
```
shaunmaharaj: I'd like to get some opinions on this message. Please take a look at my post I tagged you in on slack. We'll see what everyone suggests as the error message 😄

src/containers/ProfilePage.jsx
```diff
@@ -89,6 +95,20 @@ class ProfilePage extends React.Component {
...
+      );
+    }
+    return (
+      <div className="loader" />
+    );
+  }
+
   render() {
     const { profileData } = this.state;
     const email = profileData && profileData._emails[0]._element ? profileData._emails[0]._element[0].email : '';
```
shaunmaharaj:
```suggestion
    const email = profileData && profileData._emails && profileData._emails[0]._element ? profileData._emails[0]._element[0].email : '';
```

shaunmaharaj: There are multiple checks that need to be made here. `profileData ` will get returned but it will be missing sub-resources. Those need to be checked. Right now this will result in a broken page:
image

shaunmaharaj:
The user will not be able to add items to cart either. `productdisplayitem.main.jsx` needs to be updated to also check for the `addtocartform` in the product page before checking availability for areas like this:
`let availability = (productData._addtocartform[0].links.length > 0);`

shaunmaharaj:
@petro97 Please take a look at `appheader.main.jsx`. @rostyk-kanafotskyy added code there to get cart quantity for the header. We could potentially re-use this to hide the cart button altogether if it's not available. Then we don't have to worry about displaying an error message on the cart page at all. Same should be applicable for Wish List.

src/components/appheader.main.jsx
```diff
@@ -124,35 +127,38 @@ class AppHeaderMain extends React.Component {
 
           <div className="central-container">
             <div className="horizontal-menu">
+              {availability && (
```
shaunmaharaj: We shouldn't hide the navigation altogether here. The `CATALOG_BROWSER` role has access to browse the catalog, but cannot add to cart or wishlists. So on the Product Display Page, we should disable those buttons if the forms are not returned by the API.

src/components/appheader.main.jsx
```diff
@@ -124,35 +127,38 @@ class AppHeaderMain extends React.Component {
...
-              data-toggle="collapse"
-              data-target=".collapsable-container"
-              aria-expanded="false"
-              aria-label="Toggle navigation"
-              onClick={this.handleInputFocus}
-            >
-              <div className="search-icon" />
-            </button>
-          </div>
+          {availability && (
```
shaunmaharaj: We shouldn't hide the search altogether here. The `CATALOG_BROWSER` role has access to browse the catalog, but cannot add to cart or wishlists. So on the Product Display Page, we should disable those buttons if the forms are not returned by the API.

src/components/appheader.main.jsx
```diff
@@ -219,7 +225,9 @@ class AppHeaderMain extends React.Component {
           <hr className="mobile-navigation-separator" />
 
           <div className="mobile-navigation-container">
+            {availability && (
```
shaunmaharaj: We shouldn't hide the navigation altogether here. The `CATALOG_BROWSER` role has access to browse the catalog, but cannot add to cart or wishlists. So on the Product Display Page, we should disable those buttons if the forms are not returned by the API.

src/components/appheader.main.jsx
```diff
@@ -103,6 +105,7 @@ class AppHeaderMain extends React.Component {
     const {
       isOffline, cartData, isLoading, isSearchFocused,
     } = this.state;
+    const availability = !!cartData;
```
shaunmaharaj: cartData should not control if other elements on the page are viewable. cartData will only control if the cart can be reached. There are multiple resources for various other features. We should rename this variable to `cartAvailability` to not cause confusion
example: defaultProfile will determine if the profile data is available.

src/components/appheader.main.jsx
```diff
@@ -187,7 +193,7 @@ class AppHeaderMain extends React.Component {
         </div>
 
         <div className="collapsable-container collapse collapsed">
-          {(!Config.b2b.enable || (Config.b2b.enable && isB2bCartSelected)) && (
+          {(!Config.b2b.enable || (Config.b2b.enable && isB2bCartSelected && availability)) && (
```
shaunmaharaj: We shouldn't hide the search altogether here. The CATALOG_BROWSER role has access to browse the catalog, but cannot add to cart or wishlists. So on the Product Display Page, we should disable those buttons if the forms are not returned by the API.

src/containers/WishListsPage.jsx
```diff
@@ -93,6 +100,24 @@ class WishListsPage extends React.Component {
     });
   }
 
+  checkPermissions() {
```
shaunmaharaj: Is this still needed now that you're hiding wishlists from the top menu?

src/containers/ProfilePage.jsx
```diff
@@ -89,9 +96,24 @@ class ProfilePage extends React.Component {
...
+    }
+    return (
+      <div className="loader" />
+    );
+  }
+
   render() {
     const { profileData } = this.state;
-    const email = profileData && profileData._emails[0]._element ? profileData._emails[0]._element[0].email : '';
+    const email = profileData && profileData._emails && profileData._emails[0]._element ? profileData._emails[0]._element[0].email : '';
```
shaunmaharaj: The profile page is still currently broken with this implementation. We need to check if resources are available in the API before displaying them.

PR #263, opened 2019-03-08
Fix console errors in browser

shaunmaharaj:
Description:

- Fix errors in console
- Update Parallax and use new properties for new config
- Add keys where needed
- Add default props where needed

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/src/components/appheadertop.main.jsx
+++ b/src/components/appheadertop.main.jsx
@@ -39,7 +39,7 @@ class AppHeaderTop extends React.Component {
     const { isMobileView } = this.props;
 
     return [
-      <div key={`AppHeaderTop_${Math.random().toString(36).substr(2, 9)}`} className={`top-header ${isMobileView ? 'mobile-view' : ''}`}>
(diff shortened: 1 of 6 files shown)


src/components/appheadertop.main.jsx
```diff
@@ -28,15 +28,18 @@ import AppHeaderLocaleMain from './appheaderlocale.main';
...
+  static defaultProps = {
+    isMobileView: false,
+  }
 
   render() {
     const { isMobileView } = this.props;
 
     return [
-      <div className={`top-header ${isMobileView ? 'mobile-view' : ''}`}>
+      <div key={`AppHeaderTop_${Math.random().toString(36).substr(2, 9)}`} className={`top-header ${isMobileView ? 'mobile-view' : ''}`}>
```
skarpyak: it is better to move as util function:
getUniqueKey(key) {
 return `${key}_${Math.random().toString(36).substr(2, 9)}`
}

shaunmaharaj: Thanks @skarpyak! I changed this to a static key since there wasn't a need for it to be unique on the page.

PR #297, opened 2019-05-02
Added error message, product without price, to the quick order form

petro97:
Description:

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:


src/localization/en-CA.json
```diff
@@ -24,6 +24,7 @@
   "sku-invalid-message": "is invalid. Please clear SKU to submit order.",
   "bulk-order-invalid-message": "Some codes of products are invalid. Please review all codes.",
   "configurable-product-message": "{SKUCode} is a configurable product. Please visit the product page to add it to the cart.",
+  "product-message-without-price": "{SKUCode} does not have a price.",
```
shaunmaharaj: Was there an error message returned by the API that can be used for this? Our intention is to display those messages if they're available.

petro97: I check price field before adding to the card, after item lookup, so that’s why I hardcoded this message

PR #299, opened 2019-05-06
Only displays region if address contains one

maennis-ep:
Description:
In the AddressContainer and ProfileAddressesMain it is assumed that there will be a region to display in the address.  This fix removes the `null, `, or empty commas values that are displayed 

Linting:
- [x] No linting errors

Tests:
- [x] Manual tests
1. Create a new address in a country with no regions (ex. Great Britain)
2. Navigate to `/profile` and verify that the address their no longer contains a `null, `
3. Navigate to `/checkout` and verify that the address no longer contains an empty set of commas.

Documentation:

### DIFF
--- a/src/components/address.container.jsx
+++ b/src/components/address.container.jsx
@@ -43,10 +43,13 @@ const AddressContainer = (props) => {
           {address.locality}
           ,&nbsp;
         </span>
-        <span className="address-region">
-          {address.region}
-          ,&nbsp;
-        </span>
+        {(address.region)
+          ? (
+            <span className="address-region">
+              {address.region}
+              ,&nbsp;
+            </span>
+          ) : ('')}
         <span className="address-country">
           {address['country-name']}
           &nbsp;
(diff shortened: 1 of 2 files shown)


src/components/profileaddresses.main.jsx
```diff
@@ -89,9 +89,13 @@ class ProfileAddressesMain extends React.Component {
                       <span className="address-city" data-el-value="address.city">
                         {`${address.locality}, `}
                       </span>
-                      <span className="address-region" data-el-value="address.region">
-                        {`${address.region}, `}
-                      </span>
+                      {(address.region)
```
shaunmaharaj: Thanks for the fix! Was it intentional to remove the surrounding span tags?

src/components/address.container.jsx
```diff
@@ -43,10 +43,13 @@ const AddressContainer = (props) => {
           {address.locality}
           ,&nbsp;
         </span>
-        <span className="address-region">
-          {address.region}
-          ,&nbsp;
-        </span>
+        {(address.region)
```
shaunmaharaj: Thanks for the fix! Was it intentional to remove the surrounding span tags?

maennis-ep: I did not think about that, fix pushed!

PR #319, opened 2019-06-13
Fix caching policy and remove tokens on 401

shaunmaharaj:
Description:

- Removing tokens will force a tokens call to handle cases where cached tokens expire
- Need to update policy for service worker to not cache ANY HTTP responses

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/src/components/appheadernavigation.main.jsx
+++ b/src/components/appheadernavigation.main.jsx
@@ -25,7 +25,7 @@ import ReactRouterPropTypes from 'react-router-prop-types';
 import { withRouter } from 'react-router';
 import { Link } from 'react-router-dom';
 import _ from 'lodash';
-import { login, logout } from '../utils/AuthService';
+import { login } from '../utils/AuthService';
 import { cortexFetch } from '../utils/Cortex';
 
 import './appheadernavigation.main.less';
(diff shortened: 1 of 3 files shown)


src/utils/Cortex.js
```diff
@@ -46,6 +46,18 @@ export function cortexFetch(input, init) {
 
   return fetch(`${Config.cortexApi.path}${input}${queryFormat}`, requestInit)
     .then((res) => {
+      if (res.status === 401 || res.status === 403) {
```
shaunmaharaj: @petro97 I know this is a bit ugly but until we have our Cortex lib, we can catch all un-authenticated responses here and clear localStorage in order to fetch new tokens

src/components/appheadernavigation.main.jsx
```diff
@@ -137,18 +136,6 @@ class AppHeaderNavigationMain extends React.Component {
             Authorization: localStorage.getItem(`${Config.cortexApi.scope}_oAuthToken`),
           },
         }))
-      .then((res) => {
```
shaunmaharaj: @petro97 I'm doing some clean up here since we no longer need this code if it's in Cortex.js

PR #346, opened 2019-07-15
[RS-414] - RS 430 mock listloadmore

aChanEP:
Description:

Mocks the `productlistloadmore.tsx` component.  The data used to mock this component was gathered through the search page of CRA.

#347 is PR'ed against this branch.  Merge #347 before merging this branch.

Linting:

- [x] No linting errors

Tests:

- [ ] Manual tests

Manually tested the `productlistloadmore` story by pressing the `Load More` button.  Notice that it no longer stalls at the spinner animation but instead resolves to a browser alert; the browser alert gives brief information on what is happening with the component.

Documentation:


components/src/ProductListLoadmore/productlistloadmore.stories.tsx
```diff
@@ -1,15 +1,22 @@
...
+import productsData from './MockHttpResponses/GET/productData_response.json';
 
 import ProductListLoadMore from './productlistloadmore';
+import { mockProductListLoadMoreFromSearchResponse } from './productlistloadmore.api.mocks';
+import { searchLookup } from '../utils/CortexLookup';
 
-function handleDataChange(){}
-function onLoadMore(){}
+function handleDataChange(updateItems){
+  alert(`handleDataChange called to update container items.  Check console to see the value that it is given.`);
```
shaunmaharaj: Is this to simulate the action of a call being made?

aChanEP: Yeah that was the idea.  To show how the component is to be used.  With the README storybook plugin it would be more clear

shaunmaharaj: Okay makes sense. Thank you for the info @aChanEP! We'll add the further storybook enhancements to our backlog.

PR #354, opened 2019-07-18
Added same code to remove expired auth information to cortexFetch as adminFetch

maennis-ep:
Description:
Currently, a Cortex token expires after 7 days, and no refresh_token is available to renew the user's session.  Once the token expires, all Cortex requests return a 401.

When I went to look at implementing a solution to remove bad credentials I noticed that the `Cortex.js#adminFetch` already had a solution to this problem, so I copied it to `cortexFetch`.  I included the `b2b` mode settings to get wiped as well just in case.

Linting:
- [ ] No linting errors
eslint is currently not working in the project

Tests:
- [x] Manual tests
1. Run the PWA locally
2. Copy the bearer token from local storage to the `Authorization` header field in Cortex studio for whatever instance you are pointed at.
3. Revoke Access for that token
4. Refresh the PWA.  You should see the page reload with a fresh PUBLIC token and no 401s.

Documentation:

### DIFF
--- a/components/src/utils/Cortex.js
+++ b/components/src/utils/Cortex.js
@@ -37,6 +37,18 @@ export function cortexFetch(input, init) {
 
   return fetch(`${Config.cortexApi.path + input}`, requestInit)
     .then((res) => {
+      if (res.status === 401 || res.status === 403) {
+        localStorage.removeItem(`${Config.cortexApi.scope}_oAuthRole`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_oAuthScope`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_oAuthToken`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_oAuthUserName`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_b2bCart`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_oAuthTokenAuthService`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_keycloakSessionState`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_keycloakCode`);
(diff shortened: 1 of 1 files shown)


shaunmaharaj:
@maennis-ep Thanks for pushing this. We had a check like this before but removed since it's planned to be handled by our cortex-client in the future. I'll merge this as a temporary measure to help with. Thanks again!

components/src/utils/Cortex.js
```diff
@@ -37,6 +37,18 @@ export function cortexFetch(input, init) {
...
+        localStorage.removeItem(`${Config.cortexApi.scope}_oAuthRole`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_oAuthScope`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_oAuthToken`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_oAuthUserName`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_b2bCart`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_oAuthTokenAuthService`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_keycloakSessionState`);
+        localStorage.removeItem(`${Config.cortexApi.scope}_keycloakCode`);
+        window.location.pathname = '/';
+        window.location.reload();
```
shaunmaharaj: I recall why we didn't add this immediately. This actually doesn't give the best experience. I'll need to revert it. 
This will also be triggered if an incorrect password is entered for login, and reloading on a 401 will cause the login modal to be dismissed upon reload.

maennis-ep: That makes sense, we don't want to affect user experience in other parts of the app.  We can add this to our deployments for now to address this issue, because that use case doesn't concern us too much.

PR #360, opened 2019-07-23
Add linting to app, components project

rostyk-kanafotskyy:
Description:

Added es lint to projects and fixed all errors and warnings
Fixes #341 
Ticket: https://jira.elasticpath.com/browse/RS-454

Linting:

- [x] No linting errors

Tests:

- [x] E2E tests (npm test run with `e2e`)
- [x] Manual tests

Documentation:

- [ ] Requires documentation updates

Issue #352, opened 2019-07-18
appModalBundleConfigurationComponent has infinite spinner when adding a dependant option to cart


aChanEP:
### Step 1: Are you in the right place?

  * For issues or feature requests related to the code **in this repository** file a Github issue.
  * For issues relating to EP-Commerce, or Cloud Ops in relation to the reference store please Open a Support ticket.

### Step 2: Describe your environment

  * Device: MacOSX
  * OS version: 10.14.5
  * Storefront pulled version: 78d7dec
  * Node version: _____
  * Docker version: _____
  * Docker Compose version: _____

#### Steps to reproduce:

  1. Add a bundled item.  In our the demo-teams vestri-v2 catalog we can use the `Clothing Bundle`
  2. navigate to the cart page
  3. click the `configure` button to configure the bundle and `add to bag` any number of available options 

#### Observed Results:

 Notice at the bottom of the module that the spinner will go on infinitely and once the the modal is closed the cart is not updated unless the page is refreshed.

#### Expected Results:

Would expect the spinner to not spin forever and the cart to update right away as soon as the modal is closed.

#### Relevant Code:

  ```
  // TODO(you): code here to reproduce the problem
  ```

PR #361, opened 2019-07-23
Rs 468 dynamicbundle bug

aChanEP:
Description:

This PR fixes the dynamic bundle component for 7.5.1.

Notice that before this PR, dynamic bundles on 7.5.1 breaks this component.  the `add to bag` will infinitely spin without the component re-rendering.  Also if the bundle was configured with dependentlineitems, clicking `remove` would have also not rerendered the page. 

A big thing to note here is the replacement of the deprecated lifecycle method `componentWillRecieveProps` in favour of the `componentDidUpdate` method.  This method was actually causing issues with `remove` functionality because the props had not properly updated before re-render.  The method was also causing a needless re-render as it would fetch for bundle data which would cause a re-render and then re-render once again once the lifecycle method returned with an updated prop.  **We should swap out this lifecycle method for all components** as it is causing needless re-renders and slowing down our components.

Linting:

- [ ] No linting errors

Tests:

- [x] Manual tests

Documentation:


components/src/AppModalBundleConfiguration/appmodalbundleconfiguration.main.tsx
```diff
@@ -94,21 +94,25 @@ class AppModalBundleConfigurationMain extends React.Component<AppModalBundleConf
...
   }
 
-  componentWillReceiveProps(nextProps) {
+
+  componentDidUpdate(prevProps) {
     const { bundleConfigurationItems } = this.props;
-    if (bundleConfigurationItems !== nextProps.bundleConfigurationItems) {
-      this.fetchDependantItemData();
-    }
+    if (JSON.stringify(bundleConfigurationItems) !== JSON.stringify(prevProps.bundleConfigurationItems)) {
```
shaunmaharaj: Would you happen to know why the object structure check is not valid here anymore? Just wondering why we didn't convert to strings to start with.

components/src/AppModalBundleConfiguration/appmodalbundleconfiguration.main.tsx
```diff
@@ -94,21 +94,25 @@ class AppModalBundleConfigurationMain extends React.Component<AppModalBundleConf
...
+        this.fetchDependantItemData(bundleConfigurationItems);
+      }
+    
   }
 
-  fetchDependantItemData() {
-    const { bundleConfigurationItems } = this.props;
+  fetchDependantItemData(bundleConfigurationItems) {
```
shaunmaharaj: Will this pass the linter when we re-add it? Just want to be sure. Here's the PR for the linter:
https://github.com/elasticpath/react-pwa-reference-storefront/pull/360

components/src/AppModalBundleConfiguration/appmodalbundleconfiguration.main.tsx
aChanEP: It was never valid in the first place.  Javascript does not do a deep comparison of objects OOTB, instead it will compare the reference of objects.  In this case it always resulted in false and caused a re-render.  The condition was never doing anything.

components/src/AppModalBundleConfiguration/appmodalbundleconfiguration.main.tsx
aChanEP: Merged the linter branch #360 into this one and ran the linter inside `appModalBundleConfiguration` directory to fix all errors.  Since #360 is still undergoing changes should merge #360 before this one.

shaunmaharaj: Yep, I was wondering why it was like that in the first place. Cool, no further comments!

shaunmaharaj:
This is the fix for #352

PR #374, opened 2019-08-02
Cybersource updates

shaunmaharaj:
Description:

- Port cybersource updates based on WIP

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/app/src/ep.config.json
+++ b/app/src/ep.config.json
@@ -109,7 +109,7 @@
     "nodatalinks": true
   },
   "creditCardTokenization": {
-    "enable": true,
+    "enable": false,
     "lambdaURI": "/cssign",
     "overrideCustomReceiptURI": "/postredirector",
     "overrideCustomCancelURI": ""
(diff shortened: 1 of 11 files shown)


shaunmaharaj:
FYI @petro97, I brought your changes here just so we have a record of them. We can make changes here as needed later to port your fixes.

app/src/ep.config.json
```diff
@@ -109,8 +109,10 @@
     "nodatalinks": true
   },
   "creditCardTokenization": {
-    "enable": false,
-    "lambdaURI": ""
+    "enable": true,
```
shaunmaharaj: @petro97 this should be false OOTB and enabled as the integration is configured

app/src/ep.config.json
```diff
@@ -109,8 +109,10 @@
...
+    "lambdaURI": "/cssign",
```
shaunmaharaj: I'll need to add documentation around this. Will update our docs this week for it.

shaunmaharaj:
@petro97 can you please confirm if there are any further changes to port over for this one? I'll create tasks for the documentation items for this as well.

PR #379, opened 2019-08-08
[WIP] - Added ClientContext and converted appheader

dusanradovanovic:
Description:
This PR is utilizing `cortex-client` library for communicating with cortex backend. It provides `client` object through `ClientContext` React context which is then available to all components. `client` object contains functions for making calls to cortex backend. In addition to providing functions library also provides TypeScript types that can be used in React state interfaces. AppHeader component is the first one to make use of the new paradigm.

Linting:

- [ ] No linting errors

Tests:

- [ ] Manual tests

Documentation:

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -22,9 +22,10 @@
 import React from 'react';
 import { BrowserRouter as Router, Switch, withRouter } from 'react-router-dom';
 import {
-  AppHeaderMain, FacebookChat, AppFooterMain, ChatComponent,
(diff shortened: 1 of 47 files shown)


app/src/App.tsx
```diff
@@ -164,13 +165,20 @@ const Root = (props) => {
   ];
 };
 
+const cortexClient = cortex.createClient({
+  serverBaseUrl: '/cortex',
```
shaunmaharaj: This should come from ep.config.json. (cortexApi.path). Although perhaps it's worth removing it from the config if it will never change too.

PR #381, opened 2019-08-08
Added keys to Root to prevent React warnings

dusanradovanovic:
Description:
Added `key` props to element within a Root of an application since all array elements need to have unique keys.

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:


shaunmaharaj:
Thank you for fixing @dusanradovanovic!!

PR #387, opened 2019-08-14
convert profile components

yulia-dnistrian:
Description:

Linting:

- [ ] No linting errors

Tests:

- [ ] Manual tests

Documentation:

### DIFF
--- a/components/src/ProfileEmailInfo/profileemailinfo.main.tsx
+++ b/components/src/ProfileEmailInfo/profileemailinfo.main.tsx
@@ -62,7 +62,7 @@ class ProfileemailinfoMain extends React.Component<ProfileemailinfoMainProps, Pr
     this.submitEmailChange = this.submitEmailChange.bind(this);
   }
 
-  componentDidMount() {
+  async componentDidMount() {
     this.client = this.context;
   }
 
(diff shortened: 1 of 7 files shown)


components/src/ProfileEmailInfo/profileemailinfo.main.tsx
```diff
@@ -72,46 +80,38 @@ class ProfileemailinfoMain extends React.Component<ProfileemailinfoMainProps, Pr
...
-        console.error(error.message);
-      });
+        },
+      },
+    );
+
+    root.defaultprofile.emails.emailform({
+      emailId: '',
+      email,
+    }).fetch({}).then(() => {
```
dusanradovanovic: Can you change promise chain (than/catch) to await with regular try/catch.

components/src/ProfileInfo/profileInfo.main.tsx
```diff
@@ -86,58 +94,22 @@ class ProfileInfoMain extends React.Component<ProfileInfoMainProps, ProfileInfoM
...
-            // eslint-disable-next-line no-console
-              console.error(error.message);
-            });
-        }).catch((error) => {
-        // eslint-disable-next-line no-console
-          console.error(error.message);
-        });
+    const { onChange } = this.props;
+    const root = await this.client.root().fetch({ defaultprofile: {} });
+    this.client.profile(root.defaultprofile.uri).update({
```
dusanradovanovic: You are missing `await` here.

components/src/ProfileInfo/profileInfo.main.tsx
```diff
@@ -86,58 +94,22 @@ class ProfileInfoMain extends React.Component<ProfileInfoMainProps, ProfileInfoM
...
+      givenName: firstName,
+      familyName: lastName,
+      profileId: '',
     });
+    onChange();
+    this.cancel();
+    onChange();
```
dusanradovanovic: You are calling onChange() twice. Is this intentional or a mistake?

yulia-dnistrian: There was some issue with updating givenName and familyName. Sometimes after calling onChange() those fields are not updated. That is why I call onChange() twice. In this case, everything works correctly.

PR #398, opened 2019-09-03
Rs 498

aChanEP:
Description:

RS-498

This PR removes unnecessary config and build files within the components directory and replaces it with functioning ones.

A new webpack config has been placed in the root of the components project in order to support its build process.  `build` directory is populated with transpiled es5 and is consumable via these options detailed in the webpack docs: https://webpack.js.org/guides/author-libraries/.

Linting:

- [ ] No linting errors

Tests:

- [ ] Manual tests

1. Included a component within the drupal project (Having issues with the react-router dependency as it is not used in drupal due to server side routing).
2. Included the node_module of the store-component within the adjacent CRA project.

A Note:  The CRA will fail due to styling imports in `index.tsx`.  But getting rid of the styling import will enable components to be loaded.  These styling issues are outside the scope of this PR.  Should fix the styling in another ticket.  We should also rethink how we want our components to accept styling.  Some good examples of component libraries can be found here: https://github.com/brillout/awesome-react-components

Documentation:

`npm run build` inside the components will build the components to be consumable.


shaunmaharaj:
@aChanEP Styles are intended to be passed to components via props. We can discuss as part of the other ticket

components/src/ProfilePaymentMethods/profilepaymentmethods.main.tsx
```diff
@@ -21,11 +21,11 @@
 
 import React from 'react';
 import Modal from 'react-responsive-modal';
-import { PaymentFormMain } from '@elasticpath/store-components';
 import { withRouter } from 'react-router';
 import { login } from '../utils/AuthService';
 import { cortexFetch } from '../utils/Cortex';
 import { getConfig, IEpConfig } from '../utils/ConfigProvider';
+import PaymentFormMain from '../PaymentForm/paymentform.main';
```
shaunmaharaj: thank you! we missed this one

Issue #330, opened 2019-06-28
Duplicate .less files and images that are not reference properly by CRA reference store and storybook project


aChanEP:
### Step 1: Are you in the right place?
**Feature Request**

### Step 2: Describe your environment
Local Development

  * Device: MacOS Mojave 
  * OS version: 10.14.5
  * Storefront pulled version: fcc04b7f2e0e7c53a4ef5a633403c9c2789df828

### Step 3: Describe the problem:
There are duplicate style files and images for the two projects, the webapp and storybook project.  One set under the path `app/src` and another under `components/src/style`.  It seems as though only the image and style files under the path `components/src/style` are being used.  For example changing the `@mainColor` css variable under `components/src/style` changes the color of the header bar in both the reference store web application as well as the component in storybook.  Furthermore, changing any of the images under `components/src/images` changes the images in both projects as well.  Having two sets of assets can make reskinning and repurposing the reference store troublesome.

#### Steps to reproduce:

  1. Change `@mainColor` css variable under `components/src/style`
  2. Start the reference store web app as well as the components

#### Observed Results:
Notice that the theme of the two applications has changed when really it should just be the component theme?

(Can try and change the images under `components/src/images` and notice that it changes the reference store images well).  

(Can also try and change the images and variables under `app/src` and notice nothing changes in the reference store).

#### Expected Results:

Expect the changes for styles and images in each of the respective projects to be used in their respecitve projects.

PR #400, opened 2019-09-06
Rs 442

aChanEP:
Description:

This ticket resolves the issue described here: https://github.com/elasticpath/react-pwa-reference-storefront/issues/330

All images have been placed into the original CRA project.  This is due to the fact that it should be the client who is consuming the components that should be providing the images.

All the top level styles and constants have been deleted inside the components folder.  The styles folder under the original CRA project is now being used.  Again it should be the client of the components who should provide the styles.

Linting:

- [ ] No linting errors

Tests:

- [ ] Manual tests

Documentation:

1.) Change the `common.less` file constants and notice changes.  (Other common.less has been deleted to avoid confusion)
2.) Try and change an image under the `app` directory and notice changes in the images that are bundled.


shaunmaharaj:
@aChanEP As part of this PR, we'll have to update the version of the components in `react-pwa-reference-storefront/components/package.json` so the updated components are pushed to npm. The current version is `1.1.8`, and should be incremented to `1.1.9`. In the future we have to automate this step though.

aChanEP:
Will have it updated in a few moments.  Thanks!

aChanEP:
updated!

aChanEP:
resolved

PR #401, opened 2019-09-06
Updated load more and search results page to use cortex lib

rostyk-kanafotskyy:
Description:

Linting:

- [ ] No linting errors

Tests:

- [ ] Manual tests

Documentation:

### DIFF
--- a/components/src/ProductListLoadmore/productlistloadmore.tsx
+++ b/components/src/ProductListLoadmore/productlistloadmore.tsx
@@ -33,7 +33,7 @@ interface ProductListLoadMoreProps{
         [key: string]: any
     },
     handleDataChange: (...args: any[]) => any,
-    itemsZoom: cortex.NavigationFetch | { [key: string]: any }
+    itemsZoom: cortex.NavigationFetch
 }
 interface ProductListLoadMoreState {
     canLoadMore: boolean,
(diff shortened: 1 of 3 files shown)


components/src/ProductListLoadmore/productlistloadmore.tsx
```diff
@@ -32,61 +33,58 @@ interface ProductListLoadMoreProps{
         [key: string]: any
     },
     handleDataChange: (...args: any[]) => any,
-    onLoadMore: (...args: any[]) => any,
+    itemsZoom: cortex.NavigationFetch | { [key: string]: any }
```
dusanradovanovic: Why is this cortex type OR the dictionary?

components/src/ProductListPagination/productlistpagination.main.tsx
```diff
@@ -52,120 +42,39 @@ class ProductListPagination extends React.Component<ProductListPaginationProps,
...
-  componentDidMount() {
-    paginationPreviousLinkVar = '';
-    paginationNextLinkVar = '';
-    searchUrlVar = true;
-    this.setState({
-      paginationPreviousLink: paginationPreviousLinkVar,
-      paginationNextLink: paginationNextLinkVar,
-      searchUrl: searchUrlVar,
-    });
+    this.state = { paginationData: paginationDataProps };
```
dusanradovanovic: This is outside of scope of this PR but we shouldn't be storing parent props in a state. That is not a valid React pattern. You don't have to fix it for this PR but we should change this at some point.

rostyk-kanafotskyy: Thank for the comment, I'll update it here so we don't need to come back to it later.

PR #413, opened 2019-09-11
Feature/bloomreach integration

aChanEP:
Description:

This PR brings in a bloomreach search suggestions component that is dynamically swapped based on the ep.config.json.  The config has been set up to point at a demo bloomreach server that has been set up for us.  It has been loaded with the latest vestri catalog so the suggestions will be correct values.  The component is the exact same as the original search component except that it takes up the entire right side of the header bar (needed more space for suggestions with long names).

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

1.  Enable bloomreach search in ep.config.json
2. Notice that the right search bar will increase in size
3. type characters and watch suggestions appear below search bar
4. navigate the suggestions with your keyboard or click items with mouse
5. notice that search will be correctly made.

NOTE: The new component is not available when app is in mobile mode.  Not sure if we want to include the suggestions in mobile mode as well.  Also this component does not yet have a story for it.  Let me know if you'd like to include both of these things in this PR or of you'd like me to include it in another

** MERGE branch #416 & #417 before merging this one ** 

Documentation:


app/src/ep.config.json
```diff
@@ -94,6 +94,38 @@
     "pageId": "<PAGE_ID>",
     "applicationId": "<APP_ID>"
   },
+  "bloomreachSearch": {
+    "enable": false,
+    "config": {
+      "baseUri": "http://core.dxpapi.com/api/v1/core/?",
```
shaunmaharaj: Are these all default values, or are they values that should be overwritten as part of the implementation? We will need to add documentation to support this as well as it's brought back into product.

components/src/AppHeader/appheader.main.tsx
```diff
@@ -255,7 +256,11 @@ class AppHeaderMain extends React.Component<AppHeaderMainProps, AppHeaderMainSta
 
           <div className="icons-header-container">
             <div className="search-container">
-              <AppHeaderSearchMain isMobileView={false} onSearchPage={onSearchPage} />
+              {Config.bloomreachSearch.enable ? (
+                <BloomreachAppHeaderSearchMain isMobileView={false} onSearchPage={onSearchPage} />
+              ) : (
+                <AppHeaderSearchMain isMobileView={false} onSearchPage={onSearchPage} />
```
shaunmaharaj: Just to confirm, we're still using EP search on mobile then?

components/src/Bloomreach/bloomreach.appheadersearch.main.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+  setSuggestionsList(element, indexOfSuggestion: number) {
+    this.suggestionListElements[indexOfSuggestion] = element;
+  }
+
+  handleChange(event) {
+    const keywords = event.target.value;
+
+    bloomreachSuggestionSearch(keywords)
+      .then((res) => {
```
shaunmaharaj: We're trying to move away from nested promises and move toward async for promises. Please take a look at some of the other PRs here as an example:
https://github.com/elasticpath/react-pwa-reference-storefront/pull/414/files

components/src/Bloomreach/bloomreach.appheadersearch.main.tsx
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    this.searchInput.current.value = '';
+    if (event != null) {
+      event.preventDefault();
+    }
+  }
+
+  liHandleKeyDown(e) {
+    if (e.keyCode === 38) {
```
shaunmaharaj: Do you still have to use keycodes for this? How did this work on mobile?

app/src/ep.config.json
aChanEP: These are values based on an Bloomreach environment that was set up for us.  So in a real implementation some of these values should be overwritten.  Sounds good I can help out with the documentation.

components/src/AppHeader/appheader.main.tsx
aChanEP: This is just one section of the implementation.  I haven't yet included bloomreach search in any of my PR's.  Just search suggestion.  Will include bloomreach search in following PR's

components/src/Bloomreach/bloomreach.appheadersearch.main.tsx
aChanEP: Sounds good I'll make the necessary changes

components/src/Bloomreach/bloomreach.appheadersearch.main.tsx
aChanEP: Adding key codes was needed to support arrow functionality with the suggestion dropdown on desktop.  Shouldn't affect mobile? Although I did just try it on my Iphone xs max and there are some issues with clicking the dropdown.  I think its an unrelated to keycodes though.

aChanEP: Yeah can confirm it has nothing to do with keycodes.  Was missing touch events needed for mobile interaction.

aChanEP: Fixed the mobile scrolling interaction here #416

components/src/Bloomreach/bloomreach.headersearch.main.less
```diff
@@ -0,0 +1,134 @@
...
+    width: 170px;
+    text-align: right;
+  
+    .search-form-br {
+      position: relative;
+      display: inline-block;
+      
+      &:focus-within {
+        .input-search-br {
+          width: 270px;
```
shaunmaharaj: The width of this search box is fairly large. I'd suggest using 170px as the max width for the box and dropdown.

components/src/Bloomreach/bloomreach.headersearch.main.less
```diff
@@ -0,0 +1,134 @@
...
+          color: #fff;
+        }
+      }
+
+      .suggestions {
+        display: none;
+        margin-top: 0;
+        max-height: 149px;
+        overflow-y: auto;
+        width: 270px;
```
shaunmaharaj: Please set this to a smaller value to keep the width down. 
ex. 170px

components/src/utils/Bloomreach.js
```diff
@@ -0,0 +1,35 @@
...
+ *
+ *     https://www.gnu.org/licenses/
+ *
+ *
+ */
+export default function bloomreachFetch(input, requestInit) {
+  return fetch(input, requestInit)
+    .then((res) => {
+      if (res.status >= 500) {
+        if (window.location.href.indexOf('/maintenance') === -1) {
```
shaunmaharaj: If this is a general fetch, can you add it to the Cortex.ts and call from there? It would also allow future calls to re-use the same function.

components/src/utils/Bloomreach.js
```diff
@@ -0,0 +1,35 @@
+/**
```
shaunmaharaj: We're using typescript now so this should be a ts file

components/src/utils/BloomreachSearchService.js
```diff
@@ -0,0 +1,69 @@
...
+      .then((res) => {
+        resolve(res);
+      })
+      .catch((error) => {
+        // eslint-disable-next-line no-console
+        console.error(error.message);
+        reject(error);
+      });
+  }));
+}
```
shaunmaharaj: missing new line

shaunmaharaj:
Need to increment package.json and package-lock.json

components/src/utils/Bloomreach.js
aChanEP: hm yeah it is general fetch call and works exactly the same as cortex fetch.  Just didn't seem appropriate using `Cortex.ts` with bloomreach related calls.  Perhaps we could rename Cortex.ts to something general like `FetchInfo.ts` which would still hold cortex 
specific fetch functions but also bloomreach ones.

components/src/utils/Bloomreach.js
aChanEP: Sounds good i can change them to .ts.  Noticed that there are two utils folders in both components and app.  The files in components are all .js files so I will change those as well.  Not sure if we are needing both folders though.  Might need to open a seperate issue for that.

components/src/Bloomreach/bloomreach.headersearch.main.less
aChanEP: I increased the width here as some of the suggestions that are coming back from bloomreach are 3 words long and dont fit within the 170px width.  Take a look at this screen shot here:  
Screen Shot 2019-09-16 at 4 11 24 PM

the 170px width would either cut off part of the suggestion or have it wrap around on the bottom which didn't look great.  Would be open to other suggestions with handling large recommendation strings in the suggestion box

aChanEP:
> Need to increment version package.json and package-lock.json

version increment can be found here: #417

app/src/ep.config.json
shaunmaharaj: @aChanEP If these are sensitive values that we have for our environment, we shouldn't be distributing them OOTB in the project. We'll need to document these and backup our own configs somewhere else for future demos, etc.

components/src/AppHeader/appheader.main.tsx
shaunmaharaj: @aChanEP Should this also be included on mobile though? `AppHeaderSearchMain` is imported on mobile below here

components/src/Bloomreach/bloomreach.headersearch.main.less
shaunmaharaj: @aChanEP They should be wrapping in the container. 

I'd prefer either of the two options:

1) Increase the width from 270px to 275px or something higher so there's no buttons peeking out from under the search bar like below:
image

2) Decrease the width to 170px and force the suggestions to text-wrap in the container as below:

image

app/src/ep.config.json
shaunmaharaj: @aChanEP Do we have the `vestri_reference` catalog setup on Bloomreach too? I'm thinking we could remove these values, and add them for our http://reference.elasticpath.com demo instance to show the integration 😄

components/src/utils/Bloomreach.js
shaunmaharaj: @aChanEP Considering the current work with the cortex client lib, maybe we can keep this for now and revisit if it's needed later. 😄

app/src/ep.config.json
aChanEP: > @aChanEP If these are sensitive values that we have for our environment, we shouldn't be distributing them OOTB in the project. We'll need to document these and backup our own configs somewhere else for future demos, etc.

Yeah that makes total sense to me.  Not quite sure what the process is for adding new documentation.  But perhaps I can begin work on a `BR` integration section that explains these configs and can go along with this PR

aChanEP: > @aChanEP Do we have the `vestri_reference` catalog setup on Bloomreach too? I'm thinking we could remove these values, and add them for our http://reference.elasticpath.com demo instance to show the integration 😄

The catalog that should be set up with bloomreach is the one found here:  `vestri-us.epdemos.com/cm/` under the store `vestri_b2c`.  It seems to be the same store as `vestri_reference` found on `http://reference.elasticpath.com` so I'm thinking it should be no problem to add the integration to the existing demo

components/src/Bloomreach/bloomreach.headersearch.main.less
aChanEP: Ok I think i will go with option 1 here.  Good catch with the icons not being fully covered.  Weirdly, seems to only be happening for me on the firefox.  No issues with styling on chrome.

aChanEP: This issue should be fixed for firefox now

aChanEP:
Have merged #416 and #417 PR's that were branched off of this branch

app/src/ep.config.json
```diff
@@ -95,6 +95,23 @@
...
```
shaunmaharaj: @aChanEP We should leave these fields empty OOTB unless we want to allow consumers of this repo to use them as well when starting the app (similar to the asset links). We'll inject them into our reference.elasticpath.com build as part of this ticket https://jira.elasticpath.com/browse/RS-553

shaunmaharaj:
@aChanEP No further comments! We can merge this once the config is updates, then we'll continue with documentation this sprint. Thanks for contributing this one!

components/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@elasticpath/store-components",
-  "version": "1.1.9",
+  "version": "1.1.10",
```
shaunmaharaj: Need to increment this version number

aChanEP: > @aChanEP No further comments! We can merge this once the config is updates, then we'll continue with documentation this sprint. Thanks for contributing this one!

Sounds good, I'll place empty values there.

aChanEP: Empty values in the config have been pushed

PR #420, opened 2019-09-17
Fixed typescript build

dusanradovanovic:
Description:
This PR fixes the issue with the TypeScript build. Prior to this TypeScript errors were going undetected due to the wrong configuration in webpack.config (not including all projects) as well as due to using of "files" properties in tsconfig.json (which limits compiling only to the files listed there). This PR includes all files back into compilation process as well as fixing all existing errors so that build works again.

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/components/package.json
+++ b/components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@elasticpath/store-components",
-  "version": "1.1.9",
+  "version": "1.1.10",
   "main": "build/cjs/index",
   "main:src": "src/index",
   "types": "src/index",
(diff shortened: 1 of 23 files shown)


components/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.1.9",
```
shaunmaharaj: Let's also increment the version number of the components in `components/package.json` and `components/package-lock.json` so they get pushed to npm

shaunmaharaj:
No further comments. Thanks @dusanradovanovic!

PR #440, opened 2019-10-02
Feature/reskinnable header icons

aChanEP:
Description:

This PR allows for the header-icons to be reskinned (`magnifying-glass.svg` & `bag-icon.svg` & `account-icon` & `bulk-cart`).  Instead of setting the SVG images as CSS backgrounds in the html elements, the SVG's are now brought in as react-components so that we can make styling changes.  In this case we utilize the ability to change the SVG color using the 'fill' attribute.  The SVG's can be automatically brought in as components since create-react-app V2 detailed here: `https://create-react-app.dev/docs/adding-images-fonts-and-files#adding-svgs`.

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

1.) Load both storybook and the original web app and notice that the icons in the header have not changed in both b2b and b2c mode.
2.) Change the variable `@mainNavigationColor` in `common.less` and notice the icons change color.

Documentation:


shaunmaharaj:
@aChanEP This is really helpful, and the approach is very good. I notice you only updated for the search icon and bag icon. If we wanted to be consistent, we should apply this to others as well (bulk order icon, account icon, etc). Will you push an update for those too?

aChanEP:
> @aChanEP This is really helpful, and the approach is very good. I notice you only updated for the search icon and bag icon. If we wanted to be consistent, we should apply this to others as well (bulk order icon, account icon, etc). Will you push an update for those too?

Makes sense i'll push an update.  I think we should only wrap svg's we want to be reskinning.  So all the ones in the header (bulk order icon, account icon).  We can keep the css background approach for the others as I imagine there to be an added overhead of importing the svg and wrapping it as a react component.

aChanEP:
> @aChanEP This is really helpful, and the approach is very good. I notice you only updated for the search icon and bag icon. If we wanted to be consistent, we should apply this to others as well (bulk order icon, account icon, etc). Will you push an update for those too?

I've updated icons `bulk-cart` & `account-icon` to be wrapped as a react-component

PR #456, opened 2019-10-21
Gift Certificate not applying on the first purchase it is used

mwan-ep:
Description:

Gift Certificates applied from the reference store would not apply to purchase during the first order. Instead, the GC would apply to the cart after the purchase had already gone through.[Ticket](https://jira.elasticpath.com/browse/FLT-15)

The following changes were made:
- Ensure gift card is applied **before** submitting the purchase form.
- removes gift cards with 0 balance from localstorage
- fixes floating point numbers when subtraction occurred by rounding to 2 decimals.

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

[edit]
For instance information please ping me privately.

Steps taken:
1. Log in and purchase a gift card.
2. Check mail server for the gift card code.
3. Go add another item to cart and apply the gift card in the checkout screen.
4. Perform checkout, confirm in the purchase receipt the Gift card was listed as Authorized.
5. Confirm if gift card balance is 0, it does not appear in saved gift cards.
6. Ensure gift card subtraction of total is to 2 decimal places. (Test with a gift card balance that doesn't end in X.00) 

Documentation:

### DIFF
--- a/app/src/containers/OrderReviewPage.tsx
+++ b/app/src/containers/OrderReviewPage.tsx
@@ -176,75 +176,84 @@ class OrderReviewPage extends React.Component<RouteComponentProps, OrderReviewPa
     const { orderData } = this.state;
     const { history } = this.props;
     const purchaseform = orderData._order[0]._purchaseform[0].links.find(link => link.rel === 'submitorderaction').uri;
-    login().then(() => {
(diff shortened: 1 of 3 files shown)


shaunmaharaj:
Thank you for this one @mwan-ep!

@yulia-dnistrian @petro97 Can you please take a look at this one since you worked on the initial implementation? I've added you both as code reviewers for this.

PR #461, opened 2019-10-23
Fix am logout

maennis-ep:
Description:
Currently when using a B2B enabled Reference Storefront there is a bug on logout that prevents proper logout from keycloak because a page redirect happens before the call to keycloak can finish.  After logging out, clicking the Login link will prompt the user with the Shop for dialog instead of taking them to the keycloak authentication page.

Linting:
- [x] No linting errors

Component Updates:

Tests:
- [x] Manual tests
Log in and out of the front-end a dozen times and it works.  If you need access to a keycloak server and a remote AM instance please let me know and I can provide credentials for testing.

Documentation:

### DIFF
--- a/app/src/utils/AuthService.ts
+++ b/app/src/utils/AuthService.ts
@@ -205,6 +205,13 @@ export function logout() {
   }));
 }
 
+export function logoutAccountManagementUser() {
+  logout().then(() => {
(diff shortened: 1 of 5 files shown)


shaunmaharaj:
@rostyk-kanafotskyy Can you please take a look at this one as review? I'm going to merge as it's needed, but more as an FYI of this implementation.

PR #492, opened 2019-11-20
Fixed multi cart checkout issue

rostyk-kanafotskyy:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/app/src/containers/CheckoutPage.tsx
+++ b/app/src/containers/CheckoutPage.tsx
@@ -45,32 +45,38 @@ const zoomArrayProfile = [
 
 const zoomArray = [
   // zooms for checkout summary
-  'defaultcart',
-  'defaultcart:total',
-  'defaultcart:discount',
-  'defaultcart:order',
-  'defaultcart:order:tax',
-  'defaultcart:order:total',
-  'defaultcart:appliedpromotions:element',
-  'defaultcart:order:couponinfo:coupon',
-  'defaultcart:order:couponinfo:couponform',
+  'total',
+  'discount',
+  'order',
+  'order:tax',
+  'order:total',
+  'appliedpromotions:element',
+  'order:couponinfo:coupon',
+  'order:couponinfo:couponform',
   // zooms for billing address
-  'defaultcart:order:billingaddressinfo:billingaddress',
-  'defaultcart:order:billingaddressinfo:selector:choice',
-  'defaultcart:order:billingaddressinfo:selector:choice:description',
+  'order:billingaddressinfo:billingaddress',
+  'order:billingaddressinfo:selector:choice',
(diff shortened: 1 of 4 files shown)


app/src/containers/CheckoutPage.tsx
```diff
@@ -158,8 +164,11 @@ class CheckoutPage extends React.Component<RouteComponentProps, CheckoutPageStat
   }
 
   fetchOrderData() {
+    const { match } = this.props;
+    const cartCode = match.params.cart;
+    const cart = cartCode || 'default';
     login().then(() => {
-      cortexFetch(`/?zoom=${zoomArray.sort().join()}`,
+      cortexFetch(`/carts/${Config.cortexApi.scope}/${cart}?zoom=${zoomArray.sort().join()}`,
```
shaunmaharaj: @rostyk-kanafotskyy Unfortunately this will not work. `carts` is only available on EP 7.6 and up. It also may not even be available if multi-cart is disabled on 7.6, but `defaultcart` will be.

shaunmaharaj: Another approach could be to request `defaultcart`  and `carts` through the zoom, and check if `carts` is actually returned, then act on that.

app/src/containers/CheckoutPage.tsx
```diff
@@ -239,8 +248,10 @@ class CheckoutPage extends React.Component<RouteComponentProps, CheckoutPageStat
   }
 
   reviewOrder() {
-    const { history } = this.props;
-    history.push('/order');
+    const { history, match } = this.props;
```
shaunmaharaj: Awesome, this is perfect

app/src/containers/OrderReviewPage.tsx
```diff
@@ -94,8 +100,11 @@ class OrderReviewPage extends React.Component<RouteComponentProps, OrderReviewPa
   }
 
   fetchOrderData() {
+    const { match } = this.props;
+    const cartCode = match.params.cart;
+    const cart = cartCode || 'default';
     login().then(() => {
-      cortexFetch(`/?zoom=${zoomArray.sort().join()}`,
+      cortexFetch(`/carts/${Config.cortexApi.scope}/${cart}?zoom=${zoomArray.sort().join()}`,
```
shaunmaharaj: Same comment as above here. We could also switch out the entire url to fall back to `defaultcart` as another option.

app/src/containers/CheckoutPage.tsx
shaunmaharaj: Discussed over slack. Thinking about it more, we can merge as the resource names should be the same. We'll keep an eye on this.

Issue #495, opened 2019-11-24
B2bSideMenu component does not show up in storybook.


aChanEP:
### Step 1: Are you in the right place?

  * For issues or feature requests related to the code **in this repository** file a Github issue.
  * For issues relating to EP-Commerce, or Cloud Ops in relation to the reference store please Open a Support ticket.

`B2bSideMenu` component does not show up in storybook.

### Step 2: Describe your environment

  ```
  Issues submitted without sufficient information will be rejected.
  ```

  * Device: _____
  * OS version: _____
  * Storefront pulled version: _____
  * Node version: _____
  * Docker version: _____
  * Docker Compose version: _____

### Step 3: Describe the problem:
Within the storybook project there is no ui rendered for the B2bSideMenu component

#### Steps to reproduce:

  1. Turn on storybook
  2. Navigate to the b2bSideMenu component
  3. Notice that the B2bSideMenu component does not appear

#### Observed Results:

  * No visible ui for B2bSideMenu

#### Expected Results:

  * To see visible ui for the b2bsidemenu

#### Relevant Code:

  ```
  // TODO(you): code here to reproduce the problem
  ```

PR #506, opened 2019-11-26
fix b2bSideMenu

aChanEP:
Description:

Fixes issue #495 

Able to get the sidemenu component to show up in storybook.

Change logic in the sidemenu to use the passed down react-router location properties instead of the browser location properties.

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

1.) Check that the sidemenu shows up in storybook
2.) Check that the sidemenu still shows up in the reference store application

Documentation:


shaunmaharaj:
We'll update the components lib version at the end of the sprint to align with this.

PR #515, opened 2019-12-02
Localization debug messages

petro97:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:


app/src/localization/en-CA.json
```diff
@@ -352,5 +352,62 @@
   "backend-message-account.not.found": "No account with the given account name was found.",
   "cart-clear-msg": "You are about to remove all items from",
   "cart-clear-confirmation": "These items will be permanently removed. Continue?",
-  "remove-all-cart-items": "Remove all cart items"
+  "remove-all-cart-items": "Remove all cart items",
+  "bundle.does.not.contain.min.constituents": "Bundle does not contain the minimum number of required bundle constituents.",
+  "cart.empty": "Shopping cart is empty.",
+  "need.billing.address": "Billing address must be specified.",
```
shaunmaharaj: Ideally we should bring out the messages properties to their own files to keep these localization files cleaner. We can have new ones named `messages-en-CA.json`, etc.

components/src/utils/Cortex.js
```diff
@@ -62,9 +63,37 @@ export function cortexFetch(input, init) {
...
+              if (data.messages[i].id === 'field.invalid.minimum.value') {
+                debugMessageKey = message['invalid-value'] ? data.messages[i].id : `${data.messages[i].id}-msg-2`;
+              } else if (data.messages[i].id === 'item.insufficient.inventory') {
+                debugMessageKey = message['quantity-requested'] ? `${data.messages[i].id}-msg-2` : data.messages[i].id;
+              } else if (data.messages[i].id === 'item.not.available') {
+                debugMessageKey = message['field-value'] ? `${data.messages[i].id}-msg-2` : data.messages[i].id;
+              } else if (data.messages[i].id === 'item.not.in.store.catalog') {
+                debugMessageKey = message['field-value'] ? `${data.messages[i].id}-msg-2` : data.messages[i].id;
+              } else debugMessageKey = data.messages[i].id;
+              debugMessages = debugMessages.concat(`${intl.get(debugMessageKey, props)} \n `);
```
shaunmaharaj: This is perfect - to use the message id as the key for our messages. If a message does not exist in our properties file, we should fallback to displaying the message as returned by Cortex.

components/src/utils/Cortex.js
```diff
@@ -62,9 +63,37 @@ export function cortexFetch(input, init) {
...
               id: '',
             };
+            const props = {};
+
             let debugMessages = '';
+            let debugMessageKey = '';
             for (let i = 0; i < data.messages.length; i++) {
-              debugMessages = debugMessages.concat(`${data.messages[i]['debug-message']} \n `);
+              const message = data.messages[i].data;
+              props.fieldName = message['field-name'];
```
shaunmaharaj: Having to manage all these additional props can be time consuming for messages. Let's see if there's an easier way for this.

shaunmaharaj: Discussed option on Slack. We can keep the "override" in our properties file as a basic string. If a message does not exist, then we can fall-back to the one returned by Cortex. This eliminates us having to manage every error message.

I'd want to avoid us having to manually map every messages string replacement ourselves as these have the potential to change in the API. The message keys/IDs themselves should not, so we can depend on those.

shaunmaharaj:
Thank you for this @petro97! Looks good to me.
Please verify that e2e tests pass (You can use the dev environment we have in our pipeline to test) and remove the WIP tag, then we can merge 😄

PR #531, opened 2019-12-06
Support OpenID Connect Discovery for authenticating

petro97:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/components/src/AppModalLogin/appmodallogin.main.tsx
+++ b/components/src/AppModalLogin/appmodallogin.main.tsx
@@ -92,7 +92,7 @@ class AppModalLoginMain extends Component<AppModalLoginMainProps, AppModalLoginM
 
     const params = queryString.parse(url);
     if (params.code && oidcParameters) {
-      if (Config.b2b.openId.enable) {
+      if (Config.b2b.openId && Config.b2b.openId.enable) {
         localStorage.setItem(`${Config.cortexApi.scope}_openIdcCode`, params.code);
         localStorage.setItem(`${Config.cortexApi.scope}_openIdcSessionState`, params.session_state);
         localStorage.removeItem('OidcSecret');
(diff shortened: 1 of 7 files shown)


epmkelly:
looks good to me 😄

components/src/AppModalLogin/appmodallogin.main.tsx
```diff
@@ -87,8 +87,14 @@ class AppModalLoginMain extends Component<AppModalLoginMainProps, AppModalLoginM
...
+      if (Config.b2b.openId.enabled) {
+        localStorage.setItem(`${Config.cortexApi.scope}_openIdcCode`, params.code);
+        localStorage.setItem(`${Config.cortexApi.scope}_openIdcSessionState`, params.session_state);
+      } else {
+        localStorage.setItem(`${Config.cortexApi.scope}_keycloakCode`, params.code);
+        localStorage.setItem(`${Config.cortexApi.scope}_keycloakSessionState`, params.session_state);
+      }
+      localStorage.removeItem('OidcSecret');
       if (localStorage.getItem(`${Config.cortexApi.scope}_oAuthRole`) !== 'REGISTERED') {
         loginRegisteredAuthService(params.code, encodeURIComponent(Config.b2b.keycloak.callbackUrl), encodeURIComponent(Config.b2b.keycloak.client_id)).then((resStatus) => {
```
epmkelly: in this line we'll want to check if open-id feature is enabled, and make sure to set the parameters for the callback and client_id for that if so

components/src/utils/AuthService.js
```diff
@@ -230,8 +227,10 @@ export function logoutAccountManagementUser() {
     }));
   }
   return logout().then(() => {
-    const keycloakLogoutRedirectUrl = `${Config.b2b.keycloak.logoutRedirectUrl}?redirect_uri=${encodeURIComponent(Config.b2b.keycloak.callbackUrl)}`;
-    window.location.href = keycloakLogoutRedirectUrl;
+    if (!Config.b2b.openId.enable) {
```
epmkelly: i don't think we need this if statement, the above one returns something, so we shouldn't make it down here, unless this statement is already true 😄

components/src/utils/AuthService.js
```diff
@@ -215,10 +215,7 @@ export function logoutAccountManagementUser() {
           Authorization: localStorage.getItem(`${Config.cortexApi.scope}_oAuthToken`),
         },
       })
-        .then((res) => {
-          logout();
-          resolve(res);
-        })
+        .then(res => logout().then(() => resolve(res)))
```
epmkelly: awesome!

PR #536, opened 2019-12-10
Fix/fix component build

aChanEP:
Description:

https://jira.elasticpath.com/browse/RS-636

This PR aims to decrease our main `app` ref store start up time.  

In lieu of changes made to our `types` path to enable our published `build` bundle to be consumed properly we ran into the issue of not being to consume our components without building our entire `build` folder with a `webpack` command.  This PR maintains the changes made to our `types` path and a dependency to the `types` within the `build` folder, but instead of building the entire bundle with our `yarn start` command we only bulid the necessary types to run our main ref store application.

We use the `tsc` command from the `typescript` dependency.  It has been added to the root of the project as devDependency.

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

1.) Able to run `yarn start` from root and have application start up with no errors even with the `build` folder deleted before hand.

NOTE: unfamiliar with the prod deployment process.  That still needs to be tested.  Wouldn't mind seeing how thats done :).

Documentation:

### DIFF
--- a/components/tsconfig.json
+++ b/components/tsconfig.json
@@ -21,7 +21,7 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "noEmit": false,
-    "outDir": "build",
+    "outDir": "build/types",
     "esModuleInterop":true,
     "declarationMap": true
   },
--- a/package.json
+++ b/package.json
@@ -28,8 +28,9 @@
     "typescript": "^3.6.3"
   },
   "scripts": {
-    "start": "yarn build-components && cd app && yarn start",
+    "start": "yarn build-component-types && cd app && yarn start",
     "build": "yarn build-components && cd app && yarn build",
+    "build-component-types": "cd components && tsc --EmitDeclarationOnly",
     "build-components" : "cd components && yarn build",
     "test": " cd app && yarn test",
     "storybook": " cd storybook && yarn storybook",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -24,9 +24,13 @@
...
     "styled-components": "^4.3.2"
   },
+  "devDependencies": {
+    "typescript": "^3.6.3"
+  },
   "scripts": {
-    "start": "yarn build-components && cd app && yarn start",
+    "start": "yarn build-component-types && cd app && yarn start",
     "build": "yarn build-components && cd app && yarn build",
+    "build-component-types": "cd components && tsc --EmitDeclarationOnly",
```
dusanradovanovic: If this does the trick do we need the full `build-components` command in `build` script?

aChanEP: Yeah we could just use the `tsc` command and transpile everything to es5 with the types declarations into `build`.  Should just work for CRA consumers of components.

aChanEP: so have `"build-components": cd components && tsc` and keep `build-component-types` as is.

dusanradovanovic: No, my mistake, I mixed up commands, build-components run something else :) In any case i think we can merge this.

components/tsconfig.json
```diff
@@ -21,7 +21,7 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "noEmit": false,
-    "outDir": "build",
+    "outDir": "build/types",
```
shaunmaharaj: As discussed, we'll publish this later

PR #546, opened 2019-12-21
Removed monorepo

dusanradovanovic:
Description:

This PR removes yarn workspaces and lerna monorepo configuration from the project. Application is now consuming all components directly "from code" while components folder continues to have it's own package.json file so that it can be published to npm.

This is a rather large change with many files being moved so please make sure to fetch this and test it on your machines.

In order to build/publish components you must run `yarn install` in components folder and then run `yarn build` and finally `yarn publish` in the same folder.

Linting:

- [ ] No linting errors

Component Updates:

Tests:

- [ ] Manual tests

Documentation:


package.monorepo.json
```diff
@@ -0,0 +1,39 @@
+{
```
aChanEP: Do we need this file?  I think this is a legacy package.json that should be deleted?

aChanEP:
storybook does not seem to start

aChanEP:
We should also delete all things relating to lerna?  I see that the lerna config file is still in the project as well as its dependency in package.json

package.json
```diff
@@ -1,39 +1,146 @@
...
-      "**/babel-loader"
-    ]
-  },
   "dependencies": {
-    "aws-amplify": "^1.1.29",
-    "aws-amplify-react": "^2.3.9",
-    "core-js": "^2.6.9",
-    "lerna": "^3.14.1",
+    "@babel/core": "^7.7.7",
+    "@elasticpath/store-components": "^1.1.48",
```
aChanEP: We should remove the dependancy to the `@elasticpath/store-components` now that we are consuming the components as code?

package.monorepo.json
dusanradovanovic: You are right, I've deleted it.

package.json
dusanradovanovic: Removed

dusanradovanovic:
Fixed the storybook issue. @aChanEP Let me know if you can run it now.

aChanEP:
> Fixed the storybook issue. @aChanEP Let me know if you can run it now.

I am able to run both storybook and the main application.  I'm still seeing the lerna dependency though.

package.json
```diff
@@ -1,39 +1,145 @@
...
-    "test": " cd app && yarn test",
-    "storybook": " cd storybook && yarn storybook",
-    "deploy": "cd app && yarn deploy"
+    "start": "BROWSER=none node scripts/start.js",
+    "build": "node scripts/build.js",
+    "build-components" : "cd src/components && yarn build",
+    "test": "node scripts/test.js src/test",
+    "lint": "eslint ./src/**/*.js --max-warnings=0 --format=codeframe",
+    "deploy": "aws s3 sync build/ s3://ref-store-new",
+    "storybook": "export NODE_OPTIONS=--max_old_space_size=2048 && cd storybook && start-storybook -p 9009",
```
shaunmaharaj: After rebase, need to add to this cmd:
` "storybook": "export NODE_OPTIONS=--max_old_space_size=2048 && cd storybook && start-storybook -p 9009 -s ./.storybook/images",`

package.json
```diff
@@ -1,39 +1,145 @@
...
   },
   "scripts": {
-    "start": "yarn build-component-types && cd app && yarn start",
-    "build": "yarn build-components && cd app && yarn build",
-    "build-component-types": "cd components && tsc --EmitDeclarationOnly",
-    "build-components" : "cd components && yarn build",
-    "test": " cd app && yarn test",
-    "storybook": " cd storybook && yarn storybook",
-    "deploy": "cd app && yarn deploy"
+    "start": "BROWSER=none node scripts/start.js",
```
shaunmaharaj: While we're here, let's see if we can omit `BROWSER=none` from this command. It's unix only so would break on Windows systems running the app.

shaunmaharaj:
I can make the necessary documentation updates for this before we merge, once it's out of WIP

shaunmaharaj:
@dusanradovanovic I'm still seeing an error with storybook since `storybook/package.json` is no longer present:

```
nfo @storybook/react v5.2.8
info 
info => Loading presets
info => Loading presets
info => Loading custom addons config.
info => Loading custom Webpack config (full-control mode).
info => Loading create-react-app config.
ERR! Error: Cannot find module '/Users/smaharaj/git/react-pwa-reference-storefront/storybook/package.json'
ERR!     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:580:15)
ERR!     at Function.Module._load (internal/modules/cjs/loader.js:506:25)
ERR!     at Module.require (internal/modules/cjs/loader.js:636:17)
ERR!     at require (internal/modules/cjs/helpers.js:20:18)
ERR!     at getPublicUrl (/Users/smaharaj/git/react-pwa-reference-storefront/node_modules/@react-workspaces/react-scripts/config/paths.js:34:19)
```

src/ep.config.json
```diff
@@ -2,7 +2,7 @@
   "cortexApi": {
     "path": "/cortex",
     "scope": "vestri",
-    "pathForProxy": "http://localhost:9080",
+    "pathForProxy": "https://reference.epdemos.com",
```
shaunmaharaj: This should be:

`"pathForProxy": "http://localhost:9080",`

package.json
dusanradovanovic: Has anyone using Windows complained about it? We should advise anyone using Windows to use WSL. It's a built in feature and has been part of a Windows 10 for a few years now. It's a go to for any cross platform development since it has a full bash compatibility.

I personally prefer not to run a browser at startup since it opens a page in my default browser which I'm not using for development.

package.json
dusanradovanovic: 👍

package.json
shaunmaharaj: @dusanradovanovic We'll chat over Slack. There was a post that came in.

PR #547, opened 2019-12-24
Added page for redirection after login with open ID

petro97:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -133,10 +133,11 @@ const VersionContainer = (props) => {
   );
 };
 
-const Root = () => {
+const Root = (props) => {
+  const { componentsData } = props;
   const { error } = React.useContext(ErrorContext);
   return [
-    <VersionContainer key="version-container" appVersion={packageJson.version} />,
+    <VersionContainer key="version-container" componentsVersion={componentsData.version} appVersion={packageJson.version} />,
     <FacebookChat key="facebook-chat" config={Config.facebook} handleFbAsyncInit={handleFbAsyncInit} />,
     <AppHeaderMain
       key="app-header"
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -20,7 +20,9 @@
  */
 
 import React from 'react';
-import { BrowserRouter as Router, Switch, withRouter } from 'react-router-dom';
+import {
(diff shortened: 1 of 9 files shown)


app/src/App.tsx
```diff
@@ -130,11 +133,10 @@ const VersionContainer = (props) => {
   );
 };
 
-const Root = (props) => {
-  const { componentsData } = props;
+const Root = () => {
   const { error } = React.useContext(ErrorContext);
   return [
-    <VersionContainer key="version-container" componentsVersion={componentsData.version} appVersion={packageJson.version} />,
```
shaunmaharaj: why are we removing this?

app/src/App.tsx
```diff
@@ -158,7 +160,7 @@ const Root = (props) => {
     <div key="app-content" className="app-content">
       <Switch>
         {routes.map(route => (
-          <RouteWithSubRoutes key={route.path} {...route} />
+          <RouteWithSubRoutes key={`${route.path}_${Math.random().toString(36).substr(2, 9)}`} {...route} />
```
shaunmaharaj: thanks for fixing!!

app/src/App.tsx
shaunmaharaj: Ah, now I see why you removed this. I'll add a suggested snippet further below 😉

app/src/App.tsx
```diff
@@ -174,7 +176,10 @@ const AppWithRouter = (props) => {
     <Router>
       <ErrorDisplayBoundary>
         <CountProvider>
-          <App componentsData={componentsData} />
+          <Switch>
+            <Route path="/loggedin" exact component={LoginRedirectPage} />
+            <Route path="/" exact={false} component={App} componentsData={componentsData} />
```
shaunmaharaj: This doesn't actually pass the componentsData prop to the component unfortunately. Router doesn't seem to pass them 😞 

Here's an alternative you can do which does using render:
```suggestion
            <Route path="/" exact={false} render={passed_props => <App {...passed_props} componentsData={componentsData} />} />
```

petro97: Thank you @shaunmaharaj

PR #553, opened 2020-01-08
Create UI skeleton of requisition list

yulia-dnistrian:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:


components/src/AppHeader/appheader.main.tsx
```diff
@@ -289,6 +313,18 @@ class AppHeaderMain extends Component<AppHeaderMainProps, AppHeaderMainState> {
             </Link>
           </div>
 
+          <div className="links-container">
```
shaunmaharaj: We may need to revisit this after getting more influence from @BonnieEP. There are some updates expected for the navigation header that we need to consider.

app/src/ep.config.json
```diff
@@ -2,7 +2,7 @@
   "cortexApi": {
     "path": "/cortex",
     "scope": "vestri",
-    "pathForProxy": "http://localhost:9080",
+    "pathForProxy": "http://10.11.12.22:9080",
```
shaunmaharaj: Hardcoded environment values

app/src/ep.config.json
```diff
@@ -31,11 +31,11 @@
   "siteImagesUrl": "https://elasticpath-demo-images.s3-us-west-2.amazonaws.com/VESTRI_VIRTUAL_V1/%fileName%",
   "enableOfflineMode": false,
   "b2b": {
-    "enable": false,
-    "accounts": false,
+    "enable": true,
+    "accounts": true,
```
shaunmaharaj: We'll need to add a config to only show requisition lists in the dashboard view if it's set to true. Similar to how we had for "accounts" previously, we just need one for "req-list" as this is intended to be dark code.

components/src/AppHeader/appheader.main.tsx
```diff
@@ -289,6 +313,18 @@ class AppHeaderMain extends Component<AppHeaderMainProps, AppHeaderMainState> {
...
+            <RequisitionListsLink />
```
shaunmaharaj: This shows up even if a user is not logged in yet. We'll need to hide it until login. We can discuss with @BonnieEP on the best approach/location of this as we're planning to update the navigation as well for b2b.

rajomic:
a couple of thoughts:
- in a list details, remove the "product options" column. 
- I'll first speak to bonnie, but I think the bulk-edit thing needs some minor re-thinking.

BonnieEP:
@rajomic what's the thinking behind removing the product options?

PR #561, opened 2020-01-14
Rs-644 update styles

yulia-dnistrian:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:


shaunmaharaj:
@yulia-dnistrian can we update the general color palette to match what @BonnieEP had in her mockups? Notice these buttons still use the old colors
image

src/theme/common.less
```diff
@@ -20,7 +20,7 @@
  */
 
 
- @mainColor: #040060;
+ @mainColor: #000000;
```
shaunmaharaj: I just noticed that changing this will also force the background-color in src/theme/styles.less as well.

We'll need that value to be white to ensure it matches the header. I've attached a screenshot of what it looks like presently:
image

src/containers/HomePage.tsx
```diff
@@ -53,89 +53,150 @@ const HomePage: React.FunctionComponent = () => {
...
+                <img className="parallax-image" alt="home-espot-1" src={Config.siteImagesUrl.replace('%fileName%', homeEspotParallax1FileName)} onError={(e: any) => { e.target.src = homeEspot3; }} />
+              </div>
+            </Parallax>
+            <div className="sub-section">
+              <div className="text-block">
+                <span className="line line-white">
+                  {intl.get('home-sub-espot-container2-second-line')}
+                </span>
+                <span className="line line-link">
                 Learn more
```
shaunmaharaj: I know this has been here already, but can we please replace it with the localized string version? We shouldn't have hardcoded strings

src/containers/HomePage.tsx
```diff
@@ -53,89 +53,150 @@ const HomePage: React.FunctionComponent = () => {
...
+                <img className="parallax-image" alt="home-espot-1" src={Config.siteImagesUrl.replace('%fileName%', homeEspotParallax2FileName)} onError={(e: any) => { e.target.src = homeEspotParallax2; }} />
+              </div>
+            </Parallax>
+            <div className="sub-section">
+              <div className="text-block">
+                <span className="line line-white">
+                  {intl.get('home-sub-espot-container2-second-line')}
+                </span>
+                <span className="line line-link">
                 Learn more
```
shaunmaharaj: I know this has been here already, but can we please replace it with the localized string version? We shouldn't have hardcoded strings

src/containers/HomePage.less
```diff
@@ -321,3 +321,111 @@
...
   }
 }
+
+.home-page-b2b {
+  .main-banner {
+    min-height: 480px;
+    display: flex;
+    align-items: flex-end;
+    margin-bottom: 45px;
+    background: url('@{imgUrl}/carousel-images/baner_1.jpg') no-repeat center center;
```
shaunmaharaj: Sent updated image over slack

src/containers/HomePage.tsx
```diff
@@ -53,89 +53,150 @@ const HomePage: React.FunctionComponent = () => {
...
-          <div className="sub-section">
-            <img className="small-image" alt="home-espot-2" src={Config.siteImagesUrl.replace('%fileName%', homeEspot2FileName)} onError={(e: any) => { e.target.src = homeEspot3; }} />
-            <div className="text-block">
-              <span className="line line-1">
-                {intl.get('home-sub-espot-container2-first-line')}
-              </span>
-              <span className="line line-2">
-                {intl.get('home-sub-espot-container2-second-line')}
-              </span>
+  if (!Config.b2b.enable) {
```
shaunmaharaj: I'm wondering if this is the best way of presenting the b2b vs b2c homepage - strictly from a code cleanliness point of view. 
What if we had sub-components that are displayed in HomePage.tsx (B2BHomePage.tsx, B2CHomePage.tsx) and switch between those based on this condition, rather than having the large blocks returned? Let me know what you think 😄

yulia-dnistrian: It is a great idea, will implement it

src/localization/en-CA.json
```diff
@@ -375,5 +375,24 @@
...
+  "service-heading": "White glove service",
+  "service-description": "You will receive a call from our delivery partner within 72 hours◊◊ after your purchase to schedule a delivery date for products in stock. In cases where the order includes one or more special-order items, you will receive a courtesy call from Customer Service – Cooks Appliances Central within 72 hours to confirm your order. When the goods are received, our delivery partner will contact you to schedule a delivery date. Please provide the following info upon confirmation of your delivery: The number of used appliances to pick-up If you purchased an espresso machine please specify whether you will need help with altering the doorway, or any other special request. You will receive an automated call on the day before the scheduled delivery date to confirm the timeframe you have been given.",
```
shaunmaharaj: Can we please remove the extra characters: ◊◊

PR #567, opened 2020-01-24
RS-633 Quick/Bulk order with Multi-cart 

petro97:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/src/components/src/BulkOrder/bulkorder.main.less
+++ b/src/components/src/BulkOrder/bulkorder.main.less
@@ -117,6 +117,11 @@
         .scan-barcode-title {
           margin-right: 5px;
         }
+
+        .form-content-submit {
+          display: inline-block;
+          margin-right: 10px;
+        }
       }
     }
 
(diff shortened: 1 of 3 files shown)


src/components/src/BulkOrder/bulkorder.main.less
```diff
@@ -117,6 +117,11 @@
         .scan-barcode-title {
           margin-right: 5px;
         }
+
+        .form-content-submit {
+          display: inline-block;
+          margin-right: 10px;
```
shaunmaharaj: This is good, but the style is missing when the user is logged out.
image

src/components/src/BulkOrder/bulkorder.main.tsx
```diff
@@ -296,15 +451,19 @@ class BulkOrder extends Component<BulkOrderProps, BulkOrderState> {
...
-                </button>
+                {multiCartData && multiCartData._carts ? (
+                  <SelectCartButton isQuickOrder />
+                ) : (
+                  <button
+                    className="ep-btn primary small btn-itemdetail-addtocart"
+                    id="add_to_cart_quick_order_button"
+                    disabled={quickOrderDisabledButton}
+                    type="submit"
+                    onClick={() => { this.addAllToCart(items, true); }}
```
shaunmaharaj: Minor comment. The "item added to cart" popup will not get triggered by calling addAllToCart. If it's easy to, we should have it display here as well for both multi-cart and single cart

PR #570, opened 2020-01-30
[WIP] Refactored Checkout page to use payment component

dusanradovanovic:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:


aChanEP:
I notice that the payment widget is not showing up when trying to checkout as an anononymous user.  

I also notice that the styling for the same widget is different on two different pages:

ie.

Screen Shot 2020-02-03 at 10 21 54 AM
Screen Shot 2020-02-03 at 10 21 48 AM

src/containers/CheckoutPage.tsx
```diff
@@ -613,40 +567,18 @@ class CheckoutPage extends React.Component<CheckoutPageProps, CheckoutPageState>
...
-          <div className="modal-lg new-payment-modal">
-            <div className="modal-content">
-              <div className="modal-header">
-                <h2 className="modal-title">
-                  {intl.get('new-payment-method')}
-                </h2>
-              </div>
-              <div className="modal-body">
-                {
-                  localStorage.getItem(`${Config.cortexApi.scope}_oAuthRole`) === 'PUBLIC' ? (
```
aChanEP: @dusanradovanovic Need to place this anonymous user check through to `ProfilePaymentMethodsMain` component.  Currently, the payment form on the checkout page will always save the payment information to the profiles resource however saving to the profile is not available for anonymous users

aChanEP: I would recommend doing the localstorage check on the `***page.tsx` file and passing the props down to the `paymentform.main.tsx` file that dictates which resource it will post payment information too.  This ensures that the components don't make an assumption as to how login information in stored.

PR #583, opened 2020-02-13
RS-594 Image fallback for various filetypes

petro97:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:


src/components/src/ImageContainer/image.container.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+function ImageContainer(props: ImageContainerProps) {
+  const { fileName, imgUrl, className } = props;
+
+  const epConfig = getConfig();
+  Config = epConfig.config;
+
+  const handleError = (e, defaultImgUrl) => {
+    const { src } = e.target;
+    const types = ['.png', '.svg', '.jpg', '.tif'];
```
plundaahl-ep: I might be wrong about this, but I feel like maybe this should live in `ep.config.json`? This would give consumers the ability to do the following without making code changes:

- Control the order in which types are checked.
- Add or remove file types.
- Disable the feature if it winds up hitting render times too hard.

Also, would it be worth filtering `types` to remove whatever initial file type is provided to `props.fileName`?

As this stands now, I'm pretty sure that if `props.fileName` is `foo.png` but the actual file is `foo.jpg`, the component will make a second unnecessary request for `foo.png`, because `.png` appears in the types list before `.jpg`.

petro97: @plundaahl-ep, if we have an image "banner-1.jpg". and the first request has status code is 400 we don't send more requests with this file type. Then we will take the next file-types in the array. If the image is not on the CDN, then we display the local one.

petro97: I think I need to update this component

plundaahl-ep: Ah! Thanks @petro97, I got thrown off by the regex and I made some incorrect assumptions on my first pass through.

I did some quick modifications of your code to strip it down to just business logic, and I ran through them in my browser console. It looks like the `handleError` function jumps directly to the initially-provided type, meaning that if the `fileName` prop is a `jpg`, it will never try `png` or `svg`.

```
let curSrc;

const handleError = () => {
  const src = curSrc;
  const types = ['.png', '.svg', '.jpg', '.tif'];
  const currentType = src.replace(/.*(?=\.)/g, '');
  const i = types.indexOf(currentType);
  if (types[i + 1]) {
    curSrc = curSrc.replace(currentType, types[i + 1]);
    console.log(curSrc);
  } else {
    console.log('default');
  }
};

curSrc = 'myimg.png';
handleError(); // myimg.svg
handleError(); // myimg.jpg
handleError(); // myimg.tif
handleError(); // default

curSrc = 'myimg.jpg';
handleError(); // myimg.tif
handleError(); // default
```

plundaahl-ep: I'm not sure if you want to avoid adding state to this, but if so, maybe something like the following would work?

```
const fallbackTypes = ['.png', '.svg', '.jpg', '.tif'];

function ImageContainer(props: ImageContainerProps) {
  const { fileName, imgUrl, className } = props;

  const initType = fileName.replace(/.*(?=\.)/g, '');
  const types = [
    initType,
    ...fallbackTypes.filter(fileType => fileType !== initType)
  ];

  const handleError = (e, defaultImgUrl) => {
    const { src } = e.target;
    const currentType = src.replace(/.*(?=\.)/g, '');
  // ...
```

Thanks - I'm sure our SCs will appreciate this feature!

petro97: Thanks @plundaahl-ep, This is actually great idea, I'll use this array filter to check all types of files

src/components/src/ImageContainer/image.container.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const {
+    fileName, imgUrl, className, isSkuImage, onLoadData,
+  } = props;
+
+  const fallbackTypes = Config.imageFileTypes.types;
+  const initType = imgUrl.replace(/.*(?=\.)/g, '');
+  const types = (fallbackTypes && fallbackTypes.length > 0) ? [
+    initType,
+    ...fallbackTypes.filter(fileType => fileType !== initType),
+  ] : [];
```
shaunmaharaj: So if there's no config present, we default to showing no images? Perhaps we should keep the existing `png` as a fallback just in case.

src/components/src/ImageContainer/image.container.tsx
```diff
@@ -0,0 +1,88 @@
...
+}
+
+function ImageContainer(props: ImageContainerProps) {
+  const epConfig = getConfig();
+  Config = epConfig.config;
+  const {
+    fileName, imgUrl, className, isSkuImage, onLoadData,
+  } = props;
+
+  const fallbackTypes = Config.imageFileTypes.types;
```
shaunmaharaj: Any reason why lines 48-53 cannot be moved to within the condition you have on line 58?

src/components/src/ImageContainer/image.container.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const initType = imgUrl.replace(/.*(?=\.)/g, '');
+  const types = (fallbackTypes && fallbackTypes.length > 0) ? [
+    initType,
+    ...fallbackTypes.filter(fileType => fileType !== initType),
+  ] : [];
+  const imageSource = isSkuImage ? Config.skuImagesUrl.replace('%sku%', fileName) : Config.siteImagesUrl.replace('%fileName%', fileName);
+
+  const handleError = (e, defaultImgUrl) => {
+    const { src } = e.target;
+    if (Config.imageFileTypes.enable) {
```
shaunmaharaj: Let's also check if the top-level config is present too
```suggestion
    if (Config.imageFileTypes && Config.imageFileTypes.enable) {
```

src/components/src/ImageContainer/README.md
```diff
@@ -0,0 +1,24 @@
+# ImageContainer
+
+#### Description
+
+Displays the image that dynamically fetching from external CDN of various file-types.
+
+Initially trying to fetch primary file type received from the prop if not found try PNG, if not found try SVG, if not found try JPG, etc.
```
shaunmaharaj: Let's update this
```suggestion
Initially trying to fetch primary file type received from the prop if not found try PNG, if not found try JPG, if not found try SVG etc.

The configuration of these file types may be managed within `ep.config.json`. If no file types are provided in the configuration, the image will be used as is provided in the URL.
```

src/components/src/ImageContainer/image.container.tsx
petro97: Thanks @shaunmaharaj, I'll move it to within the condition

PR #588, opened 2020-02-18
Requisition list updates

yulia-dnistrian:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:


src/components/src/BulkOrder/bulkorder.main.less
```diff
@@ -118,7 +118,7 @@
           margin-right: 5px;
         }
 
-        .form-content-submit {
+        .cart-selection-dropdown {
```
shaunmaharaj: changing the name causes the margin to be removed when multi-cart is not available
image

PR #595, opened 2020-02-20
Add back shopper impersonation banner

mwan-ep:
Description:

This adds back the shopper impersonation banner into the reference store as the component that used the shopper impersonation banner is no longer being used. 

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/src/components/src/AppHeader/appheader.main.less
+++ b/src/components/src/AppHeader/appheader.main.less
@@ -41,6 +41,15 @@
     fill: @mainTextColor;
   }
 
+  .impersonation-notification {
+    height: 100%;
+    background-color: #8b1e24;
+    text-align: center;
+    color: @mainNavigationColor;
+    padding: 2px 0;
+    font-weight: 700;
+  }
+
   .main-container {
     position: relative;
     min-height: 60px;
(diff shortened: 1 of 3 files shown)


src/components/src/AppHeader/appheader.main.less
```diff
@@ -41,6 +41,15 @@
     fill: @mainTextColor;
   }
 
+  .impersonation-notification {
+    height: 100%;
+    background-color: #8b1e24;
```
shaunmaharaj: Now that we're re-visiting this, where does this color code come from? It's not currently in line with our color palette. 

Will merge this for now since it's needed.

maennis-ep: shopping-impersonation-notification

Here is a picture of the banner.  I am not sure where the color originally came from, but we are totally open to changing it.  At the very least we should probably move it to common.less

mwan-ep: When the idea of having a shopper impersonation banner was introduced, the criteria was to have a red banner so it was very obvious this was shopper impersonation. I picked this shade of red myself, but open to changing the colour. I agree this should be moved to `common.less`

PR #597, opened 2020-02-21
Added support for Netlify builds

dusanradovanovic:
Description:

Added basic scripts required to enable Netlify CI/CD integration. Testing on Netlify is commented out at this moment since it was unstable.

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,8 @@
   "scripts": {
     "start": "BROWSER=none node scripts/start.js",
     "build": "node scripts/build.js",
-    "build-components" : "cd src/components && yarn build",
+    "netlifybuild": "./netlifybuild.sh",
+    "build-components": "cd src/components && yarn build",
     "test": "node scripts/test.js src/test",
     "lint": "eslint ./src/**/*.js --max-warnings=0 --format=codeframe",
     "deploy": "aws s3 sync build/ s3://ref-store-new",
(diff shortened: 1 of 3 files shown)


shaunmaharaj:
Thanks @dusanradovanovic! We'll need to update our docs to mention this as well as give guidance on how it can be setup if a consumer wanted to use. 

What scope/back-end is being populated by Netlify currently for the environments that are setup?

aChanEP:
Are there any manual tests I can do with netlify before merging?

package.json
```diff
@@ -133,6 +134,7 @@
     "eslint-config-airbnb": "^17.1.1",
     "eslint-import-resolver-webpack": "^0.11.1",
     "eslint-plugin-header": "^3.0.0",
+    "http-server": "^0.12.1",
```
aChanEP: Is this a necessary dependancy for netlify?

dusanradovanovic: It is necessary in order to run tests during build. Test running is currently commented out in `netlifybuild.sh` since it was kind of hacky and seemed unstable but I left the dependancy just in case we want to re-enable it.

dusanradovanovic:
@shaunmaharaj I've enabled redirects on Netlify side which you can see in `_redirects` file. It's their syntax that mimics what nginx did for us previously. This way we don't use any vars on Netlify side and it can be changed for different branches by editing this file. It's currently pointing to https://reference.epdemos.com.

PR #601, opened 2020-02-24
Added max-lines-per-function linting rule to project

rostyk-kanafotskyy:
Description:

Linting:

- [ ] No linting errors

Component Updates:

Tests:

- [ ] Manual tests

Documentation:


src/ep.config.json
```diff
@@ -1,8 +1,8 @@
 {
   "cortexApi": {
     "path": "/cortex",
-    "scope": "vestri",
```
aChanEP: revert to OOTB ep.config.json values.

src/utils/Cortex.ts
```diff
@@ -26,6 +26,77 @@ import { ErrorInlet } from '../components/src/utils/MessageContext';
 
 import * as Config from '../ep.config.json';
 
+function getErrorMessages(data) {
```
aChanEP: yes! more functions pls!

.eslintrc.json
```diff
@@ -19,5 +19,16 @@
...
-  }
+  },
+  "rules": {
+    "max-lines-per-function": ["error"]
+  },
+  "overrides": [
+    {
+      "files": ["src/components/**", "src/containers/**"],
+      "rules": {
+        "max-lines-per-function": ["error", 120]
```
aChanEP: I think we should use the recommended 50 lines max per function.  But omit any functions having to do with rendering.  Thoughts? @shaunmaharaj  @rostyk-kanafotskyy

Use this exemption here.
// eslint-disable-next-line max-lines-per-function

shaunmaharaj: @aChanEP That's a good point, but I believe we have functions with more than 50 lines too and it could be a pain to refactor them all due to the multiple conditions we support.

I'm good with 120 as max lines for now, and we can clean up more once the time comes to integrate Cortex JS

aChanEP: works for me! could be wasted work once the CortexJs comes in.

.eslintrc.json
```diff
@@ -19,5 +19,16 @@
...
       }
     }
-  }
+  },
+  "rules": {
+    "max-lines-per-function": ["error"]
+  },
+  "overrides": [
+    {
+      "files": ["src/components/**", "src/containers/**"],
```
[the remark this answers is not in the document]
rostyk-kanafotskyy: Great catch, fixing it 🙂 
Thank you @aChanEP

aChanEP:
FYI i will be taking over this task!

PR #624, opened 2020-03-04
Fix/add names to attributes in dataset

aChanEP:
Description:

Adds names to catalog attributes in both english and spanish.  This is the root cause of import issues that education is facing.  Check with Nagesh for more information on this issue.

https://elasticpath.atlassian.net/browse/DEVOPS-2125.

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Import this dataset and check that all catalog attributes have names in both spanish and english

Documentation:


shaunmaharaj:
Thank you for jumping in to fix this @aChanEP 👍 
We'll wait on the test confirmation before merging this

aChanEP:
> Thank you for jumping in to fix this @aChanEP 👍
> We'll wait on the test confirmation before merging this

sounds good!

PR #633, opened 2020-03-12
Add custom action in AR quick look

yulia-dnistrian:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:


src/ep.config.json
```diff
@@ -54,7 +54,7 @@
   "gaTrackingId": "",
   "arKit": {
     "enable": true,
-    "skuArImagesUrl": "https://s3.amazonaws.com/referenceexp/ar/%sku%.usdz"
+    "skuArImagesUrl": "https://s3.amazonaws.com/referenceexp/ar/%sku%.usdz#callToAction=Add%20to%20cart&checkoutTitle=%title%&checkoutSubtitle=%description%&price=%price%"
```
shaunmaharaj: Can we append this to the usdz in our code instead of at the config level? Will keep things cleaner and maintain configs much easier.

src/components/src/ProductDisplayItem/productdisplayitem.main.tsx
```diff
@@ -625,6 +640,12 @@ class ProductDisplayItemMain extends Component<ProductDisplayItemMainProps, Prod
       slidesToScroll: 1,
     };
 
+    const skuArImagesUrl = Config.arKit.skuArImagesUrl
```
shaunmaharaj: This is where we can directly append those configs that you had. Should help keep things cleaner and more dynamic for the implementation 😃

src/components/src/ProductDisplayItem/productdisplayitem.main.tsx
```diff
@@ -607,8 +615,15 @@ class ProductDisplayItemMain extends Component<ProductDisplayItemMainProps, Prod
       vrMode,
       vrFileExists,
       multiImages,
+      detailsProductData,
     } = this.state;
 
+    const productDescription = detailsProductData ? detailsProductData[0]['display-value'].split('. ', 1)[0] : '';
```
shaunmaharaj: Have you seen a case where `display-value` is undefined or null? Could an error get thrown here?

src/components/src/ProductDisplayItem/productdisplayitem.main.tsx
```diff
@@ -607,8 +615,15 @@ class ProductDisplayItemMain extends Component<ProductDisplayItemMainProps, Prod
...
+
+    const { listPrice, itemPrice } = this.extractPrice(productData);
+
+    const price = listPrice !== itemPrice ? listPrice : itemPrice;
```
shaunmaharaj: nice check! 😃

src/ep.config.json
shaunmaharaj: If we move it to our code, we'll need to externalize the "Add to cart" string into the locale files as well

src/components/src/ProductDisplayItem/productdisplayitem.main.tsx
```diff
@@ -502,6 +502,14 @@ class ProductDisplayItemMain extends Component<ProductDisplayItemMainProps, Prod
     this.setState({ detailsProductData });
   }
 
+  handleArLinkClick(e) {
+    e.currentTarget.addEventListener('message', (event) => {
+      if (event.data === '_apple_ar_quicklook_button_tapped') {
+        this.addToCart(event);
```
shaunmaharaj: Nice! Does this work for products that require configurations? Ex. If the product requires line item modifiers to be set, or some config, will this still fire?

My suggestion is to only show the "add to cart" view if the addtocartaction is available, and it's available. We can selectively pass the "calltoaction" only under the following conditions `{availability && productData._addtocartform}`. That should give us a clue to whether we can add the item to cart in this view.

src/ep.config.json
shaunmaharaj: Since the descriptions get cut off, it may be helpful to make the box larger using `customHeight=large`

src/components/src/ProductDisplayItem/productdisplayitem.main.tsx
shaunmaharaj: This isn't quite displaying the description. I'm seeing in the Netlify deploy that it's pulling all attributes. We'll only want to display the price and first sentence of the description. Currently it's showing "True", which is the first attribute returned, whereas the description is the second attribute. 

IMG_0850

PR #647, opened 2020-03-20
RS-731 Fixes in Req list

petro97:
Description:

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/src/containers/b2b/RequisitionPageMain.tsx
+++ b/src/containers/b2b/RequisitionPageMain.tsx
@@ -639,7 +639,6 @@ class RequisitionPageMain extends Component<RouteComponentProps<RequisitionPageM
                             item={product}
                             hideAvailabilityLabel
                             isTableView
-                            onRemove={() => this.loadRequisitionListData(true)}
                             key={product._item[0]._code[0].code}
                             onCheck={() => { this.handleCheck(product); }}
                             isChosen={isProductChecked(product)}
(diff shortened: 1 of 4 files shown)


src/tests/e2e/wishlist.test.js
```diff
@@ -96,7 +96,7 @@ describe('Wishlist', () => {
   test('Remove wishlist item', async () => {
     const browser = await puppeteer.launch({
       args: ['--no-sandbox'],
-      slowMo: 30,
+      slowMo: 120,
```
shaunmaharaj: This might run successfully locally, but would timeout in Jenkins since the total test timeout is relatively slow.

src/containers/b2b/RequisitionPageMain.tsx
```diff
@@ -639,7 +639,6 @@ class RequisitionPageMain extends Component<RouteComponentProps<RequisitionPageM
                             item={product}
                             hideAvailabilityLabel
                             isTableView
-                            onRemove={() => this.loadRequisitionListData(true)}
```
shaunmaharaj: Where is the new Req List data being refreshed then upon remove?

petro97: @shaunmaharaj 
in cart.lineItem component on remove item is called `handleRemoveBtnClicked()` function where called handleQuantityChange() that is the same onRemove(). So the onRemove() isn't neccesery prop for Req list

PR #651, opened 2020-03-23
Removed component initialization and started importing intl from library.

dusanradovanovic:
Description:

This is the first part of a two part refactor. In this part I removed `intl` object from `getConfig()` and started importing it in components directly from library. `getConfig()` also now returns inlined epConfig object and does not require initialization anymore.

Linting:

- [x] No linting errors

Component Updates:

Tests:

- [x] Manual tests

Documentation:


src/components/src/utils/ConfigProvider.ts
```diff
@@ -110,21 +112,10 @@ export interface IEpConfig {
...
-let config: IConfig = {
-  config: {},
-  intl: { get: str => str },
-};
-
-export function init(_config: IConfig) {
-  return new Promise((resolve) => {
-    config = Object.freeze(_config);
-    resolve({ version: packageJson.version });
-  });
```
aChanEP: Why wouldn't we want to freeze our configuration object here? It should be a constant that is immutable. Will protect against any developers accidentally changing its value and causing unexpected behaviour

src/App.tsx
```diff
@@ -170,7 +165,7 @@ const Root = (props) => {
       appHeaderTopLinks={appHeaderTopLinks}
       appModalLoginLinks={appModalLoginLinks}
     />,
-    <Messagecontainer message={error} />,
+    <Messagecontainer key="message-container" message={error} />,
```
aChanEP: nice, getting rid of those warnings

src/App.tsx
```diff
@@ -134,22 +134,17 @@ const isComplianceDeclined = localStorage.getItem(`${Config.cortexApi.scope}_Com
 
 const complianceSupportModal = showCompliance && !isComplianceAccepted && !isComplianceDeclined ? <ComplianceSupportModal onAcceptDataPolicy={handleAcceptDataPolicy} /> : '';
 
-const VersionContainer = (props) => {
-  const { appVersion, componentsVersion } = props;
-  return (
-    <div className="version" style={{ display: 'none' }}>
-      <span>{ `${intl.get('app-version')}: ${appVersion}` }</span>
-      <span>{ `${intl.get('components-version')}: ${componentsVersion}` }</span>
```
aChanEP: We still have a components version available under `/src/components/package.json`.  Are we scrapping that?  We may need to change the github prompts.  Also how will we keep track with the components up in NPM?  Or are we getting rid of the components that should be up there entirely?

src/components/src/utils/ConfigProvider.ts
aChanEP: If we decide we do want to freeze these settings, maybe we can have a getIntl() function that is exported as well.

src/App.tsx
shaunmaharaj: Agree with the above point. We still need to surface the versions of those components in the dom.

src/components/src/utils/ConfigProvider.ts
dusanradovanovic: I've added Object.freeze to config (https://github.com/elasticpath/react-pwa-reference-storefront/pull/651/files#diff-1d1e22969e225d881fdb8f6a28b77ba7R117) but I would'n worry much about this since the whole function is being moved to React context in the second part of this refactoring.

src/App.tsx
dusanradovanovic: Let's discuss this further tomorrow.

src/components/src/utils/ConfigProvider.ts
```diff
@@ -116,6 +114,6 @@ interface IConfig {
 
 export function getConfig(): IConfig {
   return {
-    config: Config,
+    config: Object.freeze(Config),
```
aChanEP: looks good!

shaunmaharaj:
Need to address conflict in PR

src/App.tsx
aChanEP: reverted

aChanEP:
> Need to address conflict in PR

addressed

PR #673, opened 2020-04-22
Feature/rs 772 782 778

aChanEP:
Description:

https://elasticpath.atlassian.net/browse/RS-778
https://elasticpath.atlassian.net/browse/RS-782

This PR is the first step in many bundle sizing refactors that will help with performance.

Specifically this PR tackles initial load time performance with:

1. Splitting up routes to be lazy loaded.  So that the initial bundle size downloaded and parsed is smaller and ultimately leads the user to hit that first uncached page faster.
(container chunks are asynchonously fetched, the page renders as soon as the routed page chunk returns).  If the application is B2c mode B2b routes are entirely not bundled.

2. Lazy load VR dependancies and code to remove warnings from aframe library and lower the initial size of dependancies on initial load.

To see performance gains head into chrome dev tools performance tab to check on differences.
WebpackBundleAnalyzer is also helpful.

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:


src/b2bRoutes.ts
```diff
@@ -0,0 +1,219 @@
...
+  const RequisitionList = import(/* webpackChunkName: "RequisitionList" */ './containers/b2b/RequisitionList');
+  const RequisitionPageMain = import(/* webpackChunkName: "RequisitionPageMain" */ './containers/b2b/RequisitionPageMain');
+
+  /** All B2C routes */
+  /** Place the most likely first-visit-pages to be fetched first! */
+  const HomePage = import(/* webpackChunkName: "HomePage" */ './containers/HomePage');
+  const CartPage = import(/* webpackChunkName: "CartPage" */ './containers/CartPage');
+  const ProductDetailPage = import(/* webpackChunkName: "ProductDetailPage" */ './containers/ProductDetailPage');
+  const ProfilePage = import(/* webpackChunkName: "ProfilePage" */ './containers/ProfilePage');
+  const CategoryPage = import(/* webpackChunkName: "CategoryPage" */ './containers/CategoryPage');
```
dusanradovanovic, line 36: We shouldn't break our base components like CategoryPage into separate chunks. This will trigger many fetches in order to just show categories/products/cart and our own components aren't so big anyways. I think the better strategy would be to separate chunks into:
- Chunk 1: Our "base" components ie. everything you need to do the most basic commerce.
- Chunk 2: Our "additional" components, like b2b and everything you don't absolutely have to have for the basic commerce.
- All large imports that are not required for Chunk 1 should be one per chunk (think aframe and other large imports).
- Webpack will automatically bundle all remaining imports into the "vendor" bundle.

This way we optimize for great UX when all you need is to visit site/make a quick purchase etc. This is critical as most people do just that especially on mobile where network might be slower. If you need b2b you are probably on desktop anyway and you have different expectations. Same goes for aframe and other integrations, if you click on a 3d model you kind of expect it will take tame to load.

Let's discuss this further on our standup tomorrow.

aChanEP: This approach may slow down that initial first pixel that the user sees because the first chunk that is loaded is larger.  Although I do agree that the number of asynchronous requests may have become too high here.  I'll revisit this and try and find a good balance between numberof asynchonous requests vs bundle size.  Hopefully I find a balance with compelling numbers.

My initial thoughts are that we bundle all routes and even components within those routes that may be likely to be on that first load.  For example more likely routes may include the homepage and PDP (someone sends a product item to show someone), less likely might be a profile page or any routes relating to the checkout flow as this indicates the user has already made a visit to the site and thus chunks have been cached in the users browser.

src/containers/HomePage.tsx
```diff
@@ -32,12 +32,16 @@ import './HomePage.less';
 const HomePage: React.FunctionComponent = () => {
   if (!Config.b2b.enable) {
     return (
-      <B2CHomePage />
+      <Suspense fallback={<div />}>
```
dusanradovanovic, line 35: We can use the ternary operator here like:
```
const HomePage: React.FunctionComponent = () => (
  <Suspense fallback={<div />}>
    {Config.b2b.enable
      ? <B2BHomePage />
      : <B2CHomePage />
    }
  </Suspense>
);
```

src/containers/RouteContainers/baseRouter.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ *     https://www.gnu.org/licenses/
+ *
+ *
+ */
+import HomePage from '../HomePage';
+import ProductDetailPage from '../ProductDetailPage';
+import CategoryPage from '../CategoryPage';
+
+export default function b2cRoutes() {
```
dusanradovanovic, line 25: Can this be a constant instead of a function?

src/App.tsx
```diff
@@ -179,6 +189,9 @@ const Root = () => {
         {routes.map(route => (
           <RouteWithSubRoutes key={route.path} {...route} />
         ))}
+        <Suspense fallback={<div />}>
+          <AdditionalRoutes />
+        </Suspense>
```
dusanradovanovic, line 194: I don't think we should be placing `Suspense` as a children of `Switch`. Feels like an anti pattern. `Routes` are the only thing that should go here.

dusanradovanovic:
@aChanEP take a look at https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules for some instructions on how to do preloading and prefetching and what is the difference between the two.

src/App.tsx
aChanEP: > @aChanEP take a look at https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules for some instructions on how to do preloading and prefetching and what is the difference between the two.

Will do

src/containers/RouteContainers/baseRouter.ts
aChanEP: altered

aChanEP:
> I left a couple of comments but I guess we can merge this as it is and leave some of the for future tasks.

Great I'll capture everything here into a the refactor JIRA ticket

PR #675, opened 2020-04-28
Convert less files to sass

dusanradovanovic:
Description:

This PR converts all `.less` styles to '.sass' styles. Two languages have very similar syntax (https://sass-lang.com/guide), however Sass is much more popular these days. It is also supported OOTB by the create-react-app which will ease the transition of "un-ejecting" from default CRA.

Linting:

- [ ] No linting errors

Tests:

- [x] Manual tests

Documentation:


package.json
```diff
@@ -5,6 +5,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/core": "^7.7.7",
+    "@babel/plugin-transform-react-jsx-development": "^7.9.0",
```
aChanEP, line 8: Do we need this?

shaunmaharaj: We're not using jsx anymore, so I don't see why we would.

shaunmaharaj:
@dusanradovanovic Documentation is queued up for deployment. Can push to prod once this is merged 👍

Thanks for the great work on this one!

package.json
dusanradovanovic: We do. I'm not sure why but after adding sass files and node-sass there is an error during compilation asking for this plugin. Luckily this will all go away once we go back to un-ejected CRA.

aChanEP: Sounds good I guess we'll just leave it in as an intermediary step.  Everything else looks good to me!

PR #677, opened 2020-04-29
RS-760 Clean up item vs offers requests in zooms

petro97:
…rrayById

Description:

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:


src/components/src/utils/CortexLookup.js
```diff
@@ -25,16 +25,7 @@ import { cortexFetch } from './Cortex';
 import Config from '../../../ep.config.json';
 
 // Array of zoom parameters to pass to Cortex for navigationForm
-const navigationFormZoomArray = [
-  'items',
```
shaunmaharaj, line 29: Thanks for the cleanup Petro!

This is the only concerning point I have. With removal of items, we would be committed to only supporting offers moving forward. items were used for older versions of EP that don't have offers.

As we move forward, we should be okay with only respecting offers, but I may consider documenting this.

PR #682, opened 2020-05-01
Lazy load b2b Bulk Order & Barcode Scanner

aChanEP:
Description:

https://elasticpath.atlassian.net/browse/RS-795

This PR makes it so that bulk order, barcode scanner and all its dependancies will not be pulled down when in B2C mode.

In B2B mode bulk order will be lazy loaded after hitting the web app for the first time.  Barcode scanner will be lazy loaded once the BulkOrder component is shown and opened from the header nav.

Our lighthouse score has increased by ~14 points and we are having initial pixel viewing times in ~4 seconds.

Here is a screenshot of performance from a netlify build in B2C mode:

Screen Shot 2020-05-01 at 3 12 43 PM

Linting:

- [ ] No linting errors

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/src/components/src/AppHeader/appheader.main.tsx
+++ b/src/components/src/AppHeader/appheader.main.tsx
@@ -19,7 +19,7 @@
  *
  */
 
-import React, { Component } from 'react';
+import React, { Component, Suspense, lazy } from 'react';
 import intl from 'react-intl-universal';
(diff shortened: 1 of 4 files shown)


shaunmaharaj:
LGTM! Thanks @aChanEP! 🚀

PR #687, opened 2020-05-04
Remove all index.ts barrel file and all b2b component imports in b2c mode

aChanEP:
Description:

https://elasticpath.atlassian.net/browse/RS-802
https://elasticpath.atlassian.net/browse/RS-790

Using direct path to components for imports instead of using a barrel index.ts file.  

Removal of this makes import consistent across all components (some components were using the barrel file to import and others relative links).

We have increased speed slightly with this PR as all b2b components are not loaded in b2c mode now with the removal of the index.ts.  Components are only loaded if the container is loaded in a route.

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:


src/containers/WishListsPage.tsx
```diff
@@ -22,11 +22,10 @@
 import React from 'react';
 import intl from 'react-intl-universal';
 import { RouteComponentProps } from 'react-router-dom';
-import { WishListMain } from '../components/src/index';
 import { login } from '../utils/AuthService';
 import { cortexFetch } from '../utils/Cortex';
 import Config from '../ep.config.json';
-
+import WishListMain from '../components/src/WishList/wishlist.main';
```
dusanradovanovic, line 28: Let's keep the json import after the components.

dusanradovanovic:
I only have a minor comment. I'm so glad this file is gone 😀 Can you please make sure that all tests are passing, just in case 😃

src/containers/WishListsPage.tsx
aChanEP: Sure thing ill make that change

aChanEP:
> I only have a minor comment. I'm so glad this file is gone 😀 Can you please make sure that all tests are passing, just in case 😃

Sure I can run local tests.  But ive always had the problem where tests fail locally but pass in jenkins

shaunmaharaj:
@aChanEP I can update the Jenkins pipeline to pull from your branch so we can run remotely. Will set that up now before we merge and post back results

aChanEP:
> back

Great thanks!

shaunmaharaj:
Tests passed! All good to merge 🚀

PR #690, opened 2020-05-05
Rs 810

aChanEP:
Description:

https://elasticpath.atlassian.net/browse/RS-810

Removes duplicate utils files that we have in both components and containers.  Also changes .js files in components/utils to .ts files.

Also make slight lazy load change to HomePage.tsx

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:


shaunmaharaj:
Thanks for cleaning these up @aChanEP! Just verified the correct file changes are kept, so we're good to merge this one 👍

PR #695, opened 2020-05-07
RS-786 Add react-axe. Fix Axe errors

petro97:
Description:

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:


public/index.html
```diff
@@ -18,7 +18,7 @@
 -->
 
 <!DOCTYPE html>
-<html>
+<html lang="en">
```
shaunmaharaj, line 21: Will we have to refresh this as the shopper selects different languages?

shaunmaharaj:
@petro97 @yulia-dnistrian This looks great!

Can you please provide some instructions in the PR comments on how you ran Axe? Did you have to apply any specific rules, etc?
Screenshots of the results?

Please also rebase this one so we can merge when ready 😄

public/index.html
yulia-dnistrian: In index.tsx file value for the lang attribute is set to the selected language.

src/index.tsx
```diff
@@ -41,12 +41,19 @@ epConfig.supportedLocales.forEach((locale) => {
...
+  const axe = require('react-axe');
+  axe(React, ReactDOM, 1000);
+}
 // localisation init
 intl.init({
   currentLocale: UserPrefs.getSelectedLocaleValue(),
   locales,
 })
   .then(() => {
+    const newLocale = UserPrefs.getSelectedLocaleValue().split('-')[0];
```
shaunmaharaj, line 55: 👍

public/index.html
shaunmaharaj: Thanks @yulia-dnistrian, I see that change now. This is great!

petro97:
React-axe is a testing application with the axe-core accessibility testing library. It is run when NODE_ENV is not equal to ‘production’. The results will show in the Chrome DevTools console.
Some Axe errors are caused by components that are imported externally and cannot be fixed.

src/index.tsx
```diff
@@ -41,12 +41,19 @@ epConfig.supportedLocales.forEach((locale) => {
   locales[locale.value] = { ...localeMessages, ...debugMessages };
 });
 
+if (process.env.NODE_ENV !== 'production') {
```
shaunmaharaj, line 44: We may re-visit this later. I'm wondering if this is enough. This would mean we're still going to surface accessibility scans in dev mode too and it's not toggled.

PR #709, opened 2020-05-15
Removed styled components dependency

rostyk-kanafotskyy:
Description:

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests
- [x] Accessibility tests (no new react-axe errors in console)

Documentation:

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "sharethis-reactjs": "^1.3.3",
     "storybook": "^5.3.14",
     "style-loader": "0.23.1",
-    "styled-components": "^4.4.1",
     "terser-webpack-plugin": "1.3.0",
     "url-loader": "2.0.0",
     "use-onclickoutside": "^0.3.1",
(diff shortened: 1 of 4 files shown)


rostyk-kanafotskyy:
@shaunmaharaj @aChanEP Seems like it needs `styled-components` lib to work 😒
https://github.com/LucasBassetti/react-simple-chatbot/issues/157

shaunmaharaj:
@aChanEP @rostyk-kanafotskyy Are we trying to remove styled-components to reduce the bundle size? Is code-splitting not enough for the chatbot component?

aChanEP:
> @aChanEP @rostyk-kanafotskyy Are we trying to remove styled-components to reduce the bundle size? Is code-splitting not enough for the chatbot component?

Yes styled components gets bundled along with the chatbot.  So removing it would make the bundle size smaller.  I don't believe the initial load times would suffer because we've code split it so that it is downloaded asynchronously, but it will hinder how quickly the chatbot is able to be interacted with.

aChanEP:
> @shaunmaharaj @aChanEP Seems like it needs `styled-components` lib to work 😒
> [LucasBassetti/react-simple-chatbot#157](https://github.com/LucasBassetti/react-simple-chatbot/issues/157)

All good! atleast now we know that we can't optimize here!  I hope they make a fix.  Thanks for looking into this :)

PR #713, opened 2020-05-18
Added easy-peasy global store

dusanradovanovic:
Description:

This PR adds global state in the form of easy-peasy store. This global state will eventually be the only source of information for all user related info like authorization token, username, language selection etc. These information will be consumed by other components through the hooks that store exports.

Since other components are fetching and storing their own user info (like auth token) global store is currently implemented to "sync up" and follow that info through the system of events on `window` object. This mechanism will eventually be removed once all components that are using network calls are converted to functional components.

This also fixes the issue where user didn't see the navigation bar upon the first visit to the site.

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/src/hooks/store.tsx
+++ b/src/hooks/store.tsx
@@ -1,0 +1,139 @@
+/**
+ * Copyright © 2019 Elastic Path Software Inc. All rights reserved.
+ *
+ * This is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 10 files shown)


src/hooks/store.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  setAuthHeader: action((state, payload) => { state.authHeader = payload; }),
+  setNetworkError: action((state, error: Error) => { state.networkError = error; }),
+  setLookupForms: action((state, lookupForms: LookupForms) => { state.lookupForms = lookupForms; }),
+
+  fetchAuthHeader: thunk(async (actions, _payload, { getState }) => {
+    try {
+      const result = await fetchAuthToken();
+
+      // If legacy code set the token while we were fetching ours abort and use their token
```
shaunmaharaj, line 92: Makes sense for our intermediate usage

PR #735, opened 2020-05-28
add image converter

yulia-dnistrian:
Description:

Example how to run the script

bash ./tools/ConvertFiles.sh 

Before running the script you need to install "imagemagick", "jxrlib", "jq"
- brew install imagemagick
- brew install jxrlib
- brew install jq

Linting:

- [x] No linting errors

Tests:

- [x] Manual tests

Documentation:

### DIFF
--- a/src/ep.config.json
+++ b/src/ep.config.json
@@ -32,7 +32,7 @@
   "enableOfflineMode": false,
   "ImageContainerSrcs": {
     "types": ["webp", "jp2", "jpg", "jxr"],
-    "sizes": [["768", "620x620"], ["1092", "490x490"], ["2800", "450x450"]]
+    "sizes": ["768", "1092", "2800"]
   },
   "b2b": {
     "enable": false,
--- a/src/ep.config.json
+++ b/src/ep.config.json
@@ -31,7 +31,7 @@
   "siteImagesUrl": "https://elasticpath-demo-images.s3-us-west-2.amazonaws.com/VESTRI_VIRTUAL_V1/%fileName%",
   "enableOfflineMode": false,
   "ImageContainerSrcs": {
-    "types": ["webp", "jp2", "jpg"],
+    "types": ["webp", "jp2", "jpg", "jxr"],
     "sizes": ["768", "1092", "2800"]
   },
   "b2b": {
--- a/src/ep.config.json
+++ b/src/ep.config.json
@@ -32,7 +32,7 @@
   "enableOfflineMode": false,
   "ImageContainerSrcs": {
     "types": ["webp", "jp2", "jpg", "jxr"],
-    "sizes": ["768", "1092", "2800"]
+    "sizes": [["768", "620x620"], ["1092", "490x490"], ["2800", "450x450"]]
   },
   "b2b": {
     "enable": false,
--- a/tools/ConvertFiles.sh
+++ b/tools/ConvertFiles.sh
@@ -1,19 +1,31 @@
 #!/bin/bash
 
-arr1=( "${@:2:$1}" ); shift "$(( $1 + 1 ))"
-arr2=( "${@:2:$1}" ); shift "$(( $1 + 1 ))"
+images=( "${@:2:$1}" ); shift "$(( $1 + 1 ))"
 
-for i in "${arr1[@]}";
+screenBreakpoints=()
+while read i; do
+  screenBreakpoints+=($i)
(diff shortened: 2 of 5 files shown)


aChanEP:
@shaunmaharaj:  Before merging we should update docs and have some information surrounding this new util file

tools/ConvertFiles.sh
```diff
@@ -0,0 +1,69 @@
...
+        if [[ $file == *.png ]]; then
+          # -strip gets rid unnecessary metadata
+          # -quality 1 - 100, specifies image quality
+          # -resize creates thumbnail like images 4096@ = 64x64 16384@ 128x128
+          convert $file -strip -quality 1 -colors 255 -resize 4096@ .converted/Placeholders/$fileName.png
+        else
+          convert $file -strip -quality 20 -resize 16384@ ./converted/png/$fileName.jpg
+        fi
+
+        # TODO: Need to make images smaller too...
```
aChanEP, line 40: Remove TODO comment here

tools/ConvertFiles.sh
```diff
@@ -0,0 +1,69 @@
...
+
+        # Conversion to Next Gen formats, using solely imageMagick defaults
+
+        ## We need to downsize every single file...
```
aChanEP, line 44: Remove comment here

tools/ConvertFiles.sh
```diff
@@ -0,0 +1,69 @@
...
+
+        for format in "${arr2[@]}";
+          do
+            if [[ $format == webp ]]; then
+              # resize and convert to webp
+              convert $file -quality 100 -resize 620x620 ./converted/webp/$fileName-768w.webp
```
aChanEP, line 50: Can we make the list of screen breakpoints and the size of the image at each breakpoint an input variable in the script?  So something like at [768w, 620x620], [1092w, 490x490], [2800w, 450x450]

aChanEP: If the parameters get too large we could even have the script open up an image config file.

src/ep.config.json
```diff
@@ -32,7 +32,7 @@
   "enableOfflineMode": false,
   "ImageContainerSrcs": {
     "types": ["webp", "jp2", "jpg", "jxr"],
-    "sizes": ["768", "1092", "2800"]
+    "sizes": [["768", "620x620"], ["1092", "490x490"], ["2800", "450x450"]]
```
aChanEP, line 35: We should not change the settings in ep.config.json because not every image will be that size at a given screen breakpoint.  Ex.  at 768w we could have an image that is 768x200 or one at 200x200.

tools/ConvertFiles.sh
```diff
@@ -37,26 +49,24 @@ for i in "${arr1[@]}";
           convert $file -strip -quality 20 -resize 16384@ ./converted/png/$fileName.jpg
         fi
 
-        # TODO: Need to make images smaller too...
-
-        # Conversion to Next Gen formats, using solely imageMagick defaults
-
-        ## We need to downsize every single file...
```
aChanEP: nice thanks for getting rid of these

src/ep.config.json
aChanEP: Instead we can just have the script take an argument with what is in the ep.config.json.  Or it its too long maybe create a seperate json file just for processing images.  Something that specifies a list of directories, breakpoints, and image sizes at each breakpoint.

shaunmaharaj:
Thanks for working on this @yulia-dnistrian. I'll add documentation for running the script to our docs once we're ready. Thank you for posting the tutorial in the PR notes.

aChanEP:
Screen Shot 2020-06-01 at 1 27 07 PM

aChanEP:
Screen Shot 2020-06-01 at 3 26 40 PM


Can we get rid of the output from the jpgxr lib?
Maybe just place an output that indicates to the user that things were successful or not if that is the case.

aChanEP:
Screen Shot 2020-06-01 at 3 31 20 PM

Can we change Placeholders to PlaceholdersPng for consitency or just have on placeholders file with all the images we converted.

aChanEP:
> Thanks for working on this @yulia-dnistrian. I'll add documentation for running the script to our docs once we're ready. Thank you for posting the tutorial in the PR notes.

We should update Bellevie with JPEGXR format (missing when conversion was done manually) and VESTRI with all the next-gen formats.  It should be simple now :).  I will go ahead and do that once this is merged.

PR #741, opened 2020-06-09
[WIP] RS-838 configured CSS chunks

rostyk-kanafotskyy:
Description:

### So here is an example of how chunks can be split up
image

So for every js file, we have CSS chunk.
It works well for separate pages as `b2b.home.page`
But in other cases it sill causing performance issues in Litehouse:

**Before:**
image

**After:**
image

As we can see the loading time increased from 0.15s to 0.45s

Linting:

- [ ] No linting errors

Tests:

- [ ] Manual tests

Documentation:


shaunmaharaj:
Thank you for capturing this @rostyk-kanafotskyy and trying this out. That's unfortunate that it will increase our load times. 
Given the time spent, if our performance is negatively impacted by this, I suggest we capture findings of this attempt within the ticket and we close this as "not fixing".

aChanEP:
> Thank you for capturing this @rostyk-kanafotskyy and trying this out. That's unfortunate that it will increase our load times.
> Given the time spent, if our performance is negatively impacted by this, I suggest we capture findings of this attempt within the ticket and we close this as "not fixing".

Agree with this.  Splitting out the CSS is a valid spike here and we may use it down the line.  I guess our app hasn't hit a sizeable amount of CSS yet to warrant chunking for every single css to each bundle.  These numbers are odd though... how did splitting out the bundles cause net size to grow by 30kb?

PR #766, opened 2020-10-09
Refactor productdisplayitem.details.tsx

plundaahl-ep:
Description:
This PR contains a set of major refactors to the `productdisplayitem.details.tsx` component which are intended to make it easier to reason about. They are geared towards making `ProductDisplayItemDetails` easier to reason about by breaking it up into sub-components, and replacing the extra `render*` functions with sub-components.

Notes:
- These changes were originally made branching off `v4.4.0`, so there may be additional changes required to bring it in-line with the current master
- I understand that this contains a huge set of changes, and this is likely to require additional discussion and changes before it's accepted. Please comment with your thoughts!

Linting:

- [x] No linting errors

Tests:

- [ ] E2E tests (npm test run with `e2e`)
- [x] Manual tests
- [ ] Accessibility tests (no new react-axe errors in console)

My manual testing consisted of verifying that the Product Display Page still displays as expected, and that the standard operations still function properly. I have not run E2E tests, as I have been unable to get them to pass running against `master`.

Documentation:

- [ ] Requires documentation updates
- [x] Requires Storybook component updates

PR #771, opened 2021-01-13
Fix/quantity skuoption bug

plundaahl-ep:
Description:

Fixes a bug on the product display page for multi-sku products. This bug was introduced in a [large refactor of productdisplayitem.details.tsx](https://github.com/elasticpath/react-pwa-reference-storefront/pull/766) which I submitted back in November. The nature of the bug is as follows:

On multi-sku product pages, when the quantity is adjusted, the currently-selected SKU option is duplicated. From what I can tell, this is due to modifying shared state. To resolve this, I have removed the shared state modification (`Array.prototype.unshift` was being called on a shared array) and instead created a new array using destructuring.

I have also added a small refactor commit, which replaces the combination of for-each loop pushing to an outside array (and then modifying that array using the current index) with a map function, which I believe is more intuitive and less error-prone.

Linting:
- [x] No linting errors

Tests:
- [x] Manual tests
- [x] Accessibility tests (no new react-axe errors in console)

To verify the bug
1. Navigate to the product page for a multi-sku product
2. Update the item quantity in any way (typing, incrementing, decrementing)
3. Observe that the number of SKU options increases

To verify the fix
1. Repeat steps 1-2 of the previous test
2. Observe that the SKU options remain unchanged

Documentation:

### DIFF
--- a/src/components/src/ProductDisplayItemDetails/productdisplayitem.details.tsx
+++ b/src/components/src/ProductDisplayItemDetails/productdisplayitem.details.tsx
@@ -699,12 +699,11 @@ function ItemDetailSkuSelection({
     return null;
   }
 
-  productData._definition[0]._options[0]._element.map((ChoiceElement, index) => {
+  productData._definition[0]._options[0]._element.forEach((ChoiceElement, index) => {
     const selectorTitle = ChoiceElement['display-name'];
     const chosenItem = ChoiceElement._selector[0]._chosen[0];
-    const arraySelectors = ChoiceElement._selector[0]._choice || [];
+    const arraySelectors = [chosenItem, ...(ChoiceElement._selector[0]._choice || [])];
 
-    arraySelectors.unshift(chosenItem);
     if (selectorTitle === 'Size') {
       arraySelectors.sort((a, b) => {
         if (sizes.indexOf(a._description[0]['display-name']) < sizes.indexOf(b._description[0]['display-name'])) {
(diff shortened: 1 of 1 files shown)


src/components/src/ProductDisplayItemDetails/productdisplayitem.details.tsx
```diff
@@ -743,20 +743,20 @@ function ItemDetailSkuSelection({
...
             id={`${(ComponentEl.displayName.includes('Color')) ? 'product_display_item_sku_guide' : 'product_display_item_size_guide'}${itemIndex || ''}`}
             onChange={handleSkuSelection}
           >
-            {ComponentEl.map(Element => (
+            {ComponentEl.options.map(Element => (
 
               <SkuSelectAxisOption
                 optionData={Element}
                 axisDisplayName={ComponentEl.displayName}
-                defaultChosen={ComponentEl.defaultChousen}
```
shaunmaharaj, line 759: Thanks for catching this one!

shaunmaharaj (approved):

src/components/src/ProductDisplayItemDetails/productdisplayitem.details.tsx
```diff
@@ -726,10 +724,12 @@ function ItemDetailSkuSelection({
         return 0;
       });
     }
-    productKindsSelection.push(arraySelectors);
-    productKindsSelection[index].displayName = selectorTitle;
-    productKindsSelection[index].defaultChousen = chosenItem._description[0]['display-name'];
-    return productKindsSelection;
+
+    return {
+      displayName: selectorTitle,
```
shaunmaharaj, line 729: MUCH cleaner!

